Issue #6, opened 2020-09-21
Add more types of resistive devices


diego-plan9:
## Description and motivation

The underlying C++ simulator has a number of has a number of resistive device types that have not yet been exposed in the Python layers, in particular:

* Difference
* LinearStep
* Vector
* Transfer
* ExpStep

Making them available would facilitate using the simulator and experimenting with its features.

## Proposed solution

Expose the remaining types of resistive devices, taking as basis the way `ConstantStep` has been exposed through the different layers.

PR #17, opened 2020-09-25
Make cuda tiles init based on existing tile

diego-plan9:
## Related issues

#6

## Description

Follow-up to #14 

Make the the `CudaXTile` instantiation always based on an existing `XTile` which is passed as the only parameter to `__init__`. This prevents isolated creation of cuda-tiles, but simplifies the codebase and avoids the creation of a temporary simulator tile.


cbjuan (approved):
Good work!

PR #19, opened 2020-10-06
Add decorators for parametrized tests

### LINKED ISSUE #7: Simplify and streamline testing of multiple tiles and devices
## Description and motivation

In a relatively large number of cases, we have tests that perform the same check but vary in regards to the tile or device that they test. Currently, each test file handles the variation in a "similar but not quite identical" way, mostly relying on `TestCase` inheritance and mixins.

For maintainability, it seems we could extract the functionality they have in common and streamline it a bit, making it easier to add new tests and update existing ones. This will become specially relevant as the number of tiles/devices/layer types evolve, for example during #6 .

## Proposed solution

There are several approaches that can be combined:
* unifying the `get_tile` variations into a factory-like method
* making use of a parametrizing library

diego-plan9:
## Related issues

Closes #7 

## Description

Revise the tests, making use of decorators for parameterising the tests instead of relying on subclasses:

```python
@parametrize_over_layers(
    layers=[Linear, Conv2d, LinearCuda, Conv2dCuda],
    tiles=[ConstantStep],
    biases=[True, False]
)
class AnalogLayerTest(ParametrizedTestCase):
    """Analog layers abstraction tests."""
    def test_foo():
        print(1)
```

will automatically create `4 kinds of layers * 1 kind of device * 2 bias variations` = 8 `TestCases`.

This is aimed to reduce the copy-paste and repetition over the tests that require checking the same properties over a number of entities, specially with the new devices we have since #14 and future ones.

It is achieved via:
* adding two new decorators, parametrising over tiles and over layers
* adding a new base class for those tests (`ParametrizedTestCase`)
* using [parameterized](https://pypi.org/project/parameterized/) as an additional test dependency

## Details

Most of the existing logic for creating specific tiles and layers have been moved to `helpers.tiles` and `helpers.layers`, and will benefit from some simplification in further PRs.

The picking of `parameterized` as the underlying library was due to it supporting both `pytest` and standard `unittest` runner directly (as well as others). This is the reason for favoring using it over the `pytest` decorators, which would make our tests only compatible with `pytest`. There are other similar libraries but that offered more features or had a slightly different scope (`ddt`, `hypothesis`) - I tried to stick with the minimal that would give support to our needs.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -344,7 +344,7 @@ contextmanager-decorators=contextlib.contextmanager
 # List of members which are set dynamically and missed by pylint inference
 # system, and so shouldn't trigger E1101 when accessed. Python regular
 # expressions are accepted.
-generated-members=
+generated-members=get_layer,get_tile,get_resistive_device,bias
 
 # Tells whether missing members accessed in mixin class should be ignored. A
 # mixin class is detected if its name ends with "mixin" (case insensitive).
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,7 @@ job_compile_common: &job_compile_common
     - cd ../..
     - VERBOSE=1 pip install -v -e .
   script:
+    - pip install "parameterized==0.7.4"
     - make pytest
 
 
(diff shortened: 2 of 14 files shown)


cbjuan (approved):
LGTM 🔧

Issue #21, opened 2020-10-08
Enhance installation in conda environments


diego-plan9:
## Description and motivation

Currently, the build system and documentation assume that the package will be installed in a pure-python environment. However, some of the steps could be simplified for conda users, taking advantage of some of the pre-packaged dependencies.

## Proposed solution

Revise the install instructions, and the cmake configuration.

PR #22, opened 2020-10-09
Add conda instructions, adjust openblas

diego-plan9:
## Related issues

#21 

## Description

* Add quick-start snippet for miniconda in the install documentation in sphinx, reorganizing the rest of quick-starts a bit in the process.
* Adjust the finding of openblas adding the miniconda default paths.

## Details

This allows for compiling the package by installing only packages directly from conda, as `pybind11` includes the right `.cmake` file in its conda version, plus `openblas` takes care of installing the compiled shared libraries as well.


filemaster (approved):

diego-plan9:
Thanks Ivan - seems we got another "ghost" travis, so merging as the real one has completed.

Issue #20, opened 2020-10-07
Support basic hardware-aware training


diego-plan9:
## Description and motivation

As part of the next round of features, we should enable hardware-aware training (albeit in a basic form that can be expanded in later iterations).

## Proposed solution

Provide an interface for applying during forward pass:
* simple additive noise
* DACs, ADCs applied during forward pass

diego-plan9:
Indeed, glad to be able to mark the issue as closed and completed! Thanks to the nice additions from @maljoras at #25 and the rest of the team in other PRs, it should now be possible to perform hardware-aware training based on the `InferenceTile`.

A [couple](https://github.com/IBM/aihwkit/blob/c6cfa743d03447ef51198eb331610664bc336fb3/examples/5_simple_layer_hardware_aware.py)  of [examples](https://github.com/IBM/aihwkit/blob/c6cfa743d03447ef51198eb331610664bc336fb3/examples/6_lenet5_hardware_aware.py) can be found in the `examples/` folder, along with the documentation. Great work!

Issue #18, opened 2020-10-05
Inference with PCM statistical models


diego-plan9:
## Description and motivation

As part of the next round of features, we should add the ability to perform inference using PCM statistical models.

## Proposed solution

* provide an abstraction of PCM statistical models
* provide convenience integrations for performing inference
* integrate into existing forward pass:
  * programming noise
  * read noise
  * drift

diego-plan9:
Hand-in-hand with #20, it is great indeed to finally be able to close the issue with the inference tile in place and the neat inclusion of the noise models and related machinery. Thanks @maljoras for the great work, and @anu-pub and @Fabio-83 for the [extremely thorough documentation](https://aihwkit.readthedocs.io/en/latest/pcm_inference.html) and example!

PR #25, opened 2020-10-16
Inference tile

maljoras:
## Related issues
#18 #20 

## Description

Add new InferenceTiles that support basif hardware-ware training and inference using a statistical noise model that was fitted by real PCM devices.  

## Details
The new `InferenceTile` does the forward pass in "analog mode" whereas the backward and update pass is perfect. This tile is used for hardware-aware training to optimize a DNN model for analog inference only. Inference can be realistically tested at specified times with drift and statistical variations in the conductance values using the conductance dependend noise model. A simple conductance to weight mapping scheme is included and a simple drift compensation is applied. All of these methods are easily extendable by sub-classing.


diego-plan9 (approved):
Wonderful, thanks @maljoras !

Issue #10, opened 2020-09-21
Add additional examples


diego-plan9:
## Description and motivation

As part of improving the documentation and describing the usage of the different features, it would be ideal to include more examples in the `examples/` directory.

maljoras:
We have added 5 additional examples for now, thus closing this issue for now. We will of course continue to provide new examples in future and will open a new issue for that as fit.

PR #26, opened 2020-10-16
Added 4th example with CNN LeNET5 network with MNIST dataset

Fabio-83:
## Related issues

#10 

## Description

Added CNN example with a LeNet5 inspired network training on MNIST dataset

## Details

By toggling the RPU_CONFIG variable, both analog or floating point networks can be explored


diego-plan9 (approved):
Wonderful, thanks @Fabiez-83 !

PR #28, opened 2020-10-16
Update tests: base class, parameterizing

diego-plan9:
## Related issues

#7

## Description

* add an `AihwkitTestCase` as a convenient way of sharing asserts and
  potentially other functions shared by several tests
* add `Inference` and `InferenceCuda` test tile helpers
* revise the parametrization of existing tests, and make use of the
  new TestCase where appropiate


filemaster (approved):
Great!

PR #33, opened 2020-10-19
More advanced hardware-aware training

maljoras:
## Related issues
#20 

## Description

More advanced hardware-aware training for analog inference with the `InferenceTile`.

## Details

- different types of weight noise can be added per mini-batch during training, that is fixed during forward/backward. Updates are done on a reference weight matrix.
- connection drop
- clipping weights


diego-plan9 (approved):
Great, thanks @maljoras !

PR #34, opened 2020-10-20
Add AnalogSequential module

diego-plan9:
## Description

Add a new `AnalogSequential` module as the counterpart of regular `nn.Sequential` with analog-specific functionality, in order to allow easy operation on functions that need to be applied to the analog children of a Module.

The main benefit is that it allows to perform actions on just the `Analog` layers contained on it - this way, now `.cuda()` can be handled correctly (by running the `super()` normally, and after it finishes invoking it specifically and only on the analog layers), and the same for `drift_analog_weights` and `program_analog_weights`. As the number of analog-specific operations and functions grow, hopefully this can be used as the main point of entry, delegating into `._apply_to_analog`.

## Details

Using this class is not meant to be mandatory, and actually one of the goals is to keep fully compatibility with `Sequential` - it is mostly a convenience for avoiding to apply functions to the analog layers directly. The documentation has been updated to reflect the manual tweaks needed if not using the new class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ The format is based on [Keep a Changelog], and this project adheres to
 * Added a new `InferenceTile` that supports basic hardware-aware training
   and inference using a statistical noise model that was fitted by real PCM
   devices. (\#25)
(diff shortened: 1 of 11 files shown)


cbjuan (approved):
👌 good job!

PR #35, opened 2020-10-20
CNN example with hardware aware behavior

Fabio-83:
## Related issues

#10

## Description

LeNet5 inspired CNN with inference loop at different times to study the impact of weights drift on network accuracy


diego-plan9 (approved):
Thanks Fabio!

PR #45, opened 2020-10-22
Remove the backward IO parameter 

maljoras:
## Related issues

None

## Description

One often wants to specify the parameters of the forward and backward in tune, ie if one sets the `inp_res` of the forward, the backward needs to be the same. To facilitate this it would be great if one  could just do 
```
rpu_config.forward.inp_res = 234
rpu_config.backward = rpu_config.forward 
``` 
which would be very expressive and short. However, up to now this is not possible because of the specialization `BackwardIOParameter`

Here I removed this because it is really not needed. The only difference between fwd/bwd is the support for bound management.

## Details
 I now just added a silent change of setting BM to zero in case of backward, instead of an runtime error and added documentation so that users are aware that BM is ignored in case of backward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,16 @@ The format is based on [Keep a Changelog], and this project adheres to
 * `Fixed` for any bug fixes.
 * `Security` in case of vulnerabilities.
 
+
+## [UNRELEASED]
+
+### Removed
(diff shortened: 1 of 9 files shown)


diego-plan9 (approved):
Thanks Malte!

PR #46, opened 2020-10-23
Self-compile googletest during build

### LINKED ISSUE #42: -D BUILD_TEST=ON fails during compilation
## Description

Currently, it is not possible to use the `BUILD_TEST_ON` flag during compilation, as it results in linking errors. The culprit seems to be that the `googletest` library is compiled using a [different `libstdc++` ABI](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html) (old vs new) than pytorch on most of the distributions and systems.

## How to reproduce

Compiling with `-DBUILD_TEST=ON`.

## Other information

As a workaround, compiling manually `googletest` in order to use the right ABI should reenable compilation of the tests.

diego-plan9:
## Related issues

Closes #42 

## Description

Instead of relying on the system-wide `googletest` library, download and compile `googletest` with the right ABI flag during the build process. It uses the same approach we use for CUB (the `ExternalProject` cmake command), although a bit more complex than CUB (as in this case we do compile it).

## Details

After compiling, the way we tell `cmake` to use _our_ compiled version of `googletest` instead of the (potentially installed) system-wide version of `googletest` is not optimal - we just append the local paths to `include_directories` and `link_directories`. It seems to be enough for `cmake` to pick the right one, but we might need to refine it eventually.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,5 +90,6 @@ if (BUILD_TEST)
 
     add_test(NAME ${test_name} COMMAND $<TARGET_FILE:${test_name}>)
     set_target_properties(${test_name} PROPERTIES FOLDER tests)
+    add_dependencies(RPU_CPU GTest)
   endforeach()
 endif()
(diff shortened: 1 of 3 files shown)


filemaster (approved):

cbjuan (approved):
🛠

PR #60, opened 2020-11-02
Pretty print the RPUConfigs

maljoras:
## Related issues

None

## Description

Printing of the `rpu_configs` to the terminal was very unreadable. In particular, all values where printed in one line, even those little use options. 

Here we changed the behavior how to print `RPUConfigs`, to exclude the default values and only print those that change from the  defaults in a nice manner. For instance, a complicated config might print like:

```
UnitCellRPUConfig(
    device=TransferCompound(
        unit_cell_devices=[
            SoftBoundsDevice(w_max=0.3, w_min=-0.3),
            SoftBoundsDevice()
        ],
        transfer_every=2,
        transfer_forward=IOParameters(inp_res=0.015625),
        transfer_update=UpdateParameters()
    ),
    forward=IOParameters(inp_res=0.015625),
    backward=IOParameters(inp_res=0.015625),
    update=UpdateParameters()
)

```  
Note that printing all available settings (the old printing behaviour) can still be achieved by `repr(rpu_config)`.  

## Details

Use a Mixin class approach to the data classes overriding the `__str__` method.

### DIFF
--- a/examples/8_simple_layer_with_tiki_taka.py
+++ b/examples/8_simple_layer_with_tiki_taka.py
@@ -64,8 +64,17 @@ rpu_config.device.transfer_forward = rpu_config.forward
 # SGD update/transfer-update will be done with stochastic pulsing.
 rpu_config.device.transfer_update = rpu_config.update
 
+# print the config (default values are omitted)
+print('\nPretty-print of non-default settings:\n')
+print(rpu_config)
+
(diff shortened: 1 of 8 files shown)


diego-plan9 (approved):
Thanks Malte!

diego-plan9:
Merging without CI check as it seems we hit some travis hiccup.

PR #62, opened 2020-11-03
Docstring and style tweaks for ReferenceUnitCell

diego-plan9:
## Related issues

#61 

## Description

Round of small tweaks to #61:
* revise docstrings, mostly indentation for proper sphinx rendering and other small style tweaks
* update `raise` clauses so all of them raise `ConfigError`, reserving `ValueError` for those that are related to checking arguments of the function directly (subtle distinction - the existing ones were checking `self` attributes, so this way we stick further to the original `ValueError` definition).
* reword changelog slightly for making it a bit more consistent with the rest of the entries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,15 +17,15 @@ The format is based on [Keep a Changelog], and this project adheres to
 ## [UNRELEASED]
 
 ### Added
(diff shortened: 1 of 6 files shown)


filemaster (approved):
🚀

PR #66, opened 2020-11-06
Cmake output tweaks

maljoras:
## Related issues

In case of some pybind configurations the following output errors and warnings were shown, which however do not interfere with the compilation in any way:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: module 'pybind11' has no attribute 'get_cmake_dir'
```
and
```
CMake Warning:
  Manually-specified variables were not used by the project:
    SKBUILD 
```

## Description
This PR silences these messages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@
 
 cmake_minimum_required(VERSION 3.18.0)
 project(aihwkit C CXX)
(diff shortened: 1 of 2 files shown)


diego-plan9 (approved):
Thanks for looking into it Malte!

PR #72, opened 2020-11-10
Example 7 correction to work with CUDA

maljoras:
## Related issues

Having different devices composing a vector device is currently not supported by the CUDA implementation, because of the auto-tuning of the update CUDA kernels. Currently, the updated kernel is shared among all devices, thus all devices need to implement the same update type.  Also, vector of vector devices are actually not possible in CUDA currently because of the same reason. 

Thus here we make a simpler example for 7 just having a vector device with the same devices.      

## Description

Changing example 7 to also work with CUDA.


diego-plan9 (approved):
Thanks Malte!

Issue #59, opened 2020-11-02
Serialization of children analog layers


diego-plan9:
## Description

Currently, if a model makes use of analog layers as _children_ layers, it seems serialization is not handled properly:

```
load_state_dict
    raise RuntimeError('Error(s) in loading state_dict for {}:\n\t{}'.format(
RuntimeError: Error(s) in loading state_dict for Model:
	Unexpected key(s) in state_dict: "analog_tile_state"
```

## How to reproduce

Attempt to load the state dict of a model that has analog children layers.

## Other information

It seems the issue is related to the way the custom `analog_tile_state` is being used in the state dicts, causing it to be appended always to the top-level instead of respecting the prefix applied to the children layers.

diego-plan9:
Closing via #74 - thanks @Fabio-83 !

PR #74, opened 2020-11-11
Fix serialization of children analog layers

Fabio-83:
## Related issues

#59 

## Description

Update to the serialization of children layers


diego-plan9 (approved):
Thanks @Fabio-83 ! With this PR, we fix the handling of the custom key that is used for analog layers, making serialization of more complex models working again (more details about the cases that were not working correctly can be found in the newly introduced tests).

Issue #65, opened 2020-11-05
ImportError: cannot import name 'AnalogSGD' from 'aihwkit.optim'


spwjr:
## Description
I tried to run the py files under  tests/ and examples/ but it will return an error: ImportError: cannot import name 'AnalogSGD' from 'aihwkit.optim'

## How to reproduce
python tests/test_utils.py

## Other information
Follow the installation guide and python version is 3.7.3 

* Pytorch version:
* Package version:0.2
* OS:MAC Catalina
* Python version: 3.7.3
* Conda version (or N/A):

diego-plan9:
Hi @spwjr ,

and thanks for reporting! It seems there might be a mismatch between the version of `aihwkit` that you are using (which I'm assuming you installed from `pip`) and the "version" of the examples that you are using (in this case, I'm assuming you are using the `master` branch from this repo).

If that is the case, the `master` branch contains code that has not been released as a stable version in pip yet - in a way, it is ahead by some commits. In particular, at #54 we introduced a more comfortable `from aihwkit.optim import AnalogSGD` way to import the optimizer instead of the previous `from aihwkit.optim.analog_sgd import AnalogSGD`.

In order to solve the mismatch, can you try either:
1. using the examples from the [`v0.2.0` tag](https://github.com/IBM/aihwkit/tree/v0.2.0) instead of the `master` branch: that tag contains the codebase (including tests and examples) at the time the 0.2.0 release was made. If you cloned the repository, `git checkout v0.2.0` would change to the right tag.
2. use the code from the `master` branch instead of the pip-installed 0.2.0 release - this involves some more steps (most notably compilation, and dealing with the pythonpath - they are hopefully outlined at [the docs](https://aihwkit.readthedocs.io/en/latest/developer_install.html)) for setting up the development environment, but would allow you to use the new features introduced after `0.2.0`.

spwjr:
Thanks Diego
I tried both 
1:switch to v0.2.0 tag and it works !
2:with a new terminal and virtual env I have tried the method described in the Development setup page
and running the example from master branch, it will show error regarding symbol can not be found,  for example
from aihwkit.simulator.rpu_base import devices
ImportError: dlopen(/ibm_test/aihwkit_env/lib/python3.7/site-packages/aihwkit/simulator/rpu_base.cpython-37m-darwin.so, 2): Symbol not found: __ZN3c104impl23ExcludeDispatchKeyGuardC1ENS_11DispatchKeyE
[...]

PR #77, opened 2020-11-16
Revise installation instructions

diego-plan9:
## Related issues

#65 

## Description

After the conversation in #65 (thanks everyone for the extremely detailed feedback), expand the installation instructions with information about:
* the `_skbuild` directory
* setting the pythonpath and removing installed version
* passing compiler flags

Plus other fixes.


filemaster (approved):
Looks better !

PR #96, opened 2020-12-08
Use the same dataset root folder in examples

diego-plan9:
## Description

Use the same dataset root folder both for the training data loader and
the test data loader. torchvision creates another folder inside the
root folder (`MNIST/`) and is able to use the same data files for both
the sets.

## Details

Hopefully this avoids downloading the same data twice.


Fabio-83 (approved):
Thanks for making this changes!

PR #97, opened 2020-12-09
Use type alias for rpu_config in layers

diego-plan9:
## Description

Use a type alias for the `rpu_config` init parameter in the layers,
reducing verbosity and making it easier to propagate to new layers.

## Details

Actually comes from a small fix - in the `AnalogConv2d` layer, we were not declaring `InferenceRPUConfig` as part of the accepted types, which in turn was causing some IDE warnings in example 06.


diego-plan9:
Self-merging even if it seems travis has not triggered.

PR #99, opened 2020-12-11
RPUCuda update

maljoras:
## Description
In this PR the RPUCuda C++ backend has been updated with a number of features and enhancements, which means also that new options in the ``rpu_config`` of many devices are available:
* **deterministic implicit bit lines** for pulsed analog update
* RAPA convolution ready (training multiple tiles for one conv layer)  
* polynomial weight noise per minibatch for hardware-aware training for  inference chips 
* worst case bound management 
* average absolute max noise management
* cycle-to-cycle PCM read noise during each analog multiply-and-accumulate
* shift management (for better SNR before softmax)
* ``directUpdate`` capability for future more complicated meta devices 
* Option to omit the bias row for hardware-aware training weight noise (``copy_last_column``)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,13 +19,17 @@ The format is based on [Keep a Changelog], and this project adheres to
 
 ### Added
 
-* Deterministic bitlines for analog training (\XX)
-* More noise types for hardware-aware training for inference (polynomial)
-* Additional bound management schemes (worst case, average max, shift)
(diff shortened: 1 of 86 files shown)


diego-plan9 (approved):
Thanks Malte, great round of improvements! Merging right away in order to have the new features in place, and I'll issue a couple of small follow-ups to cover some minor tweaks.

Issue #47, opened 2020-10-26
Add new layers: Conv1d and Conv3d


diego-plan9:
## Description and motivation

As part of widening the scope of the library, it would be great to include more analog layers that can be used in different context and use cases. One of the likely first candidates would be additional convolution layers, taking advantage of the fact that we currently already have the `AnalogConv2d` layer implemented and that they would share quite some functionality.

## Proposed solution

Using `AnalogConv2d` as a reference, implement other types of convolution layers. Note that `torch` actually uses a `_ConvNd` internal module as the basis for the convolution layers: it seems likely that the same approach would help us avoid some duplication and streamline the new layers.

diego-plan9:
Thanks Fabio for adding these two layers! After #102 and #103, the new convolution layers are in place.

PR #102, opened 2020-12-15
AnalogConv3d layer

Fabio-83:
## Related issues

#47 

## Description

Added AnalogConv3d

## Details

Applies a 3D convolution over an input signal composed of several input volumes, using an analog tile for its forward, backward and update passes.

### DIFF
--- a/src/aihwkit/nn/modules/conv.py
+++ b/src/aihwkit/nn/modules/conv.py
@@ -14,9 +14,10 @@
 
 from typing import Optional, Tuple, Union
 
-from torch import arange, cat, float64, int32, ones, Tensor
-from torch.nn import Conv2d, Unfold
-from torch.nn.modules.utils import _pair
+from torch import Tensor, arange, cat, float64, int32, ones
+from torch.nn import Conv2d, Conv3d, Unfold
+from torch.nn.functional import pad
+from torch.nn.modules.utils import _pair, _triple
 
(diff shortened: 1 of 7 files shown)


src/aihwkit/nn/modules/conv.py
```diff
@@ -167,3 +168,176 @@ def extra_repr(self) -> str:
...
+        if groups != 1:
+            raise ValueError('Only one group is supported')
+        if padding_mode != 'zeros':
+            raise ValueError('Only "zeros" padding mode is supported')
+        if dilation != 1:
+            raise ValueError('Only dilation = 1 is supported')
+
+        kernel_size = _triple(kernel_size)
+        self.in_features = (in_channels // groups) * kernel_size[0] * kernel_size[1] * \
+            kernel_size[2]
```
diego-plan9, line 243:
```suggestion
        self.in_features = ((in_channels // groups) *
                            kernel_size[0] * kernel_size[1] * kernel_size[2])  # type: ignore
```

Suggested change as `mypy` seems not to be smart enough to discover that after line 241, `kernel_size` will always be a 3-item tuple and no longer an `int`.

diego-plan9 (approved):
Great, thanks @Fabio-83 ! It's great to have more analog layers into place, which hopefully can be put to use in different contexts and domains - great work on a PR that touches quite a bit of pieces of the library :clap:

Issue #111, opened 2021-01-18
Revise Python 3.9 compatibility


diego-plan9:
## Description and motivation

As Python 3.9 has been released for some reasonable amount of time, we should ensure that the package is fully compatible with it and update the tooling (travis, etc) and other meta-information.

## Proposed solution

Test 3.9 compatibility and provide fixes as needed.

diego-plan9:
After a first look, it seems no code changes are needed for 3.9 compatibility - however, we are getting a [segfault in travis](https://travis-ci.com/github/IBM/aihwkit/jobs/473105431) in #112 that seems related to https://github.com/pybind/pybind11/pull/2657.

diego-plan9:
Currently (from https://travis-ci.com/github/IBM/aihwkit/builds/219434622):
* Python `3.9` and torch `1.7.1` still segfaults at the end of the test suite
* Python `3.9` and torch `1.8.0` does not segfault

It seems we need https://github.com/pytorch/pytorch/pull/50998 for avoiding the segmentation fault - which was not included in the `1.7.1` release. Let's decide by the time of the release (ping to #52) if we just stick to 1.8 directly.

diego-plan9:
Closing in favour of #204

PR #112, opened 2021-01-19
travis: add 3.9 and 3.6+torch 1.7.1 stages

diego-plan9:
## Related issues

#111 

## Description

Add a travis stage for Python 3.9.


diego-plan9:
Finally marking this PR out of WIP:
* `Python 3.9` and `torch>=1.8` are working - a new stage is added to ensure this combination is tested by travis
* `Python 3.9` and `torch<1.7.x` still produces a segmentation fault (b2e4465 forced [that combination](https://travis-ci.com/github/IBM/aihwkit/jobs/496411894))

It seems that this segmentation fault can only be avoided due to a new release of the torch 1.7x. line that includes the fix, or by us replicating the torch approach (vendoring `pybind11` with additional patches).

PR #125, opened 2021-02-09
RPUCuda patch (write noise) 

### LINKED ISSUE #113: Segmentation fault (core dumped)
## Description
Segmentation fault (core dumped) during rebuilding a neural network model.
This is sample code:
```python
def create_model(k):
    model = AnalogSequential(AnalogLinear(784, k), nn.Sigmoid, AnalogLinear(k,10)).cuda()
return model

x = [100,200,300]
for i in x:
    ...
    model = create_model(i)
    ...
```
When I run the above code, after the first iteration, I got the message, "segmentation fault (core dumped)".
If I replace `AnalogSequential` with `nn.Sequential` and `AnalogLinear` with `nn.Linear`, it works. So, I think PyTorch and cuda are not the cause of this issue. In addition, if I run without `cuda()`, it also works.
 

## Other information

* Pytorch version: 1.7.1
* Package version: git clone
* OS: Ubuntu 18.04.2
* Python version: 3.8
* Conda version (or N/A): 
* NVIDIA Graphic Driver: 450.51.05
* CUDA Version: 11.0
* GPU: RTX 6000

maljoras:
## Related issues

Might possibly resolve the issue reported in #113 

## Description

 RPUCuda update preparing for upcoming new features (write noise) and bugfixes:
   * update cycle init fixes for vector device
   * on the fly memory allocation for diffusion and reset

### DIFF
--- a/src/aihwkit/simulator/rpu_base_src/rpu_base_tiles.cpp
+++ b/src/aihwkit/simulator/rpu_base_src/rpu_base_tiles.cpp
@@ -679,7 +679,7 @@ void declare_rpu_tiles(py::module &m) {
             CHECK_TORCH_INPUT(d_tensor);
 
             int N = x_input.size(0); // batch
-            int d_image_size = ((d_tensor.numel()/d_tensor.size(0))/d_tensor.size(1));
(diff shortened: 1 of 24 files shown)


diego-plan9 (approved):
Great, thank Malte!

PR #129, opened 2021-02-12
Weight scaling

maljoras:
## Related issues

Added feature.

## Description

Here a conversion factor (weight scaling) is introduced that scales the weight range of the cross bar in respect to the digital output. For instance during initialization for training, the `weight_scaling_omega` factor can be used to set the absolute max of the weight matrix to some value (e.g. to the maximum of the weight range) and scale the output accordingly. This ensures that tile operates in a proper dynamic regime for a given DNN and usually improves achievable accuracy.

## Details
See  [Rasch, Gokmen & Haensch (2019)](https://arxiv.org/abs/1906.02698) for algorithmic details.


diego-plan9 (approved):
Thanks Malte!

PR #132, opened 2021-02-17
Random states sizes and TransferCompound CUDA bugfix

maljoras:
## Related issues

potentially #113 

## Description
A bug fix which fixed the following issues:
* Random states sizes could cause errors for some network sizes and batch settings (only CUDA)
* `TransferCompound` fix for update with `transfer_every` smaller than batchsize (only CUDA)

## Details

Corrected the random state vector sizes and chunk size offsets.


diego-plan9 (approved):
Thanks Malte as usual!

Issue #133, opened 2021-02-18
Include analog device presets


diego-plan9:
## Description and motivation

While using the family of `RPUConfig` devices and adjust their parameters manually offers a great deal of flexibility, for some use cases being able to start with a set of pre-defined interesting configurations would be useful. This would help the users being familiar with the different options, serve as a basis for exploring the toolkit, and allow using "well-known" configurations off the shelf for a number of purposes.

## Proposed solution

Add a number of curated analog device presets to the package, representing devices in existing literature or other specific configurations that seem relevant or interesting.

diego-plan9:
Closed via #144

PR #144, opened 2021-03-03
Device and configuration presets

maljoras:
## Related issues

#133 

## Description

Addition of device and config presets. 

## Details

Device presets (called `XXXPresetDevice`) are parameter settings of devices, implemented by inheriting the underlying device data class. They can be used to configure a tile instead of the base classes like `ConstantStepDevice`.

Config presets (called `XXXPreset`) are parameter settings of the full `rpu_config` and could be used instead of defining ones own. They might specify a particular tile model choice and analog optimizer.  For instance, 

```python 
model = AnalogLinear(4, 2, rpu_config=TikiTakaReRamSBPreset())
```

would use Tiki-taka optimizer (using two ReRAM crossbars with the `ReRamSBPresetDevice` device model) and otherwise default parameters. As before, parameter choices can be adapted by giving keyword arguments or overwriting the field values.


diego-plan9 (approved):
Great, thanks Malte!

Issue #141, opened 2021-02-26
Add visualization utilities


diego-plan9:
## Description and motivation

It would be nice to add a small visualization module that provides convenient plotting functions, which would help users exploring the different RPU configurations and other core entities.

## Proposed solution

Add plotting functions as part of a module, which would also serve as starting an "utilities" sub-package aimed at providing convenient tools and helpers on top of the core functionality.

diego-plan9:
Closed via #146

PR #146, opened 2021-03-04
Visualization

maljoras:
## Related issues
#141 

## Description

A plotting tool for visualizing the device pulse response curve for a given resistivity device configuration.  

## Details

image


diego-plan9 (approved):
Thanks @maljoras !

I added some commits on top of the one that implements the (rather fancy if I might add) visualization, in order to deal with the `matplotlib` dependency. At the moment it is optional and only installed if the user explicitly selects the `visualization` extra when installing the package (ie. `pip install aihwkit[visualization]`).

We'll decide if this is the right approach before the next release, and document it more clearly.

PR #157, opened 2021-03-10
Preparing the RPUCUDA with new features

maljoras:
Update the RPUCUDA to include:
- mixed precision device
- drift weights

Also includes a number of format changes and small fixes in the C++/CUDA library.


diego-plan9 (approved):
Thanks @maljoras !

Issue #155, opened 2021-03-10
Add weight drift


diego-plan9:
## Description and motivation

Add the possibility to drift the weights during training with a power-law (similar to decay weights).

diego-plan9:
Closed via #158

PR #158, opened 2021-03-10
Optional power-law drift during training

maljoras:
## Related issues
#155 

## Description

Enables (optional) power-law drift during analog training, which can be conductance dependent and have device-to-device variability. If present, the drift will be applied once per minibatch only, just like the optional device-to-device variable diffusion or decay. 

## Details
The drift model is based on Oh et al (2019), see https://ieeexplore.ieee.org/document/8753712


diego-plan9 (approved):

diego-plan9:
Merging as the pylint failure will be solved in an upcoming PR.

Issue #215, opened 2021-04-27
example3 test_evaluation function is not testing against the whole validation data


tz-ac:
Hello developers,

first of all thank you a lot for this amazing software. 

I am not sure if the test_evaluation function in examples/03_mnist_training.py is evaluating the validation data as intended.
I downloaded the current version and changed the EPOCHS parameter to 3 for a faster training time. The test_validation function prints that the "Number Of Images Tested = 157" although the validation data includes 10000 images. In the current implementation the validation is only done for the last image of the batch. Is this intended?

Fabio-83:
Hi @tz-ac,

Thanks, we are really glad you are enjoying this toolkit!

Thanks for bringing this issue up! 
The [DataLoader](https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader) function returns batches of the validation set of length ```len(dataset) / batch_size``` and in the ```validation_data``` loader function we have specified ```batch_size = BATCH_SIZE``` which is 64. So the 157 * 64 accounts for the 10k images of the dataset with the last batch size a bit smaller than the rest. 
You are right, given the way the examples is structured the validation is called only once at the end of the training and so only validated for the 157 images. You can change the ```batch_size``` parameter in the ```validation_data``` to 1 and that would run the validation over the whole dataset but still only once after the training for the number of epochs. 
Another option is to include the validation function in the epoch loop (leaving ```batch_size = 64```) so you get an estimate of the network performance per epoch.

tz-ac:
Hi @Fabio-83,

thanks for your reply. 
Reducing the `batch_size` of the `validation_data` is a nice way to validate all images in the validation data set. 

From your answer I understand that it is the intendent behavior to only validate one image per batch.
Can you please elaborate a bit on the reason for the for loop in line 169:
```
for i in range(len(labels)):
    image = images[i].view(1, INPUT_SIZE)
    with torch.no_grad():
        pred = model(image)
```
Why is it necessary if only the last label of the validation batch is compared in line 179?
```
validation_label = labels.cpu().numpy()[-1]
if validation_label == predicted_label:
    predicted_ok += 1
```

Fabio-83:
Hi @tz-ac,

ohh, I may have not completely grasped your point earlier on. 
I see what you mean now, yes that ```for``` loop may have some issues. 
I don't think it was intentional, thanks for pointing this out, we will update the ```test_evaluation``` function. 

How about making it more similar to what we use in some of the other examples:

```
[...]

PR #217, opened 2021-04-28
Update to examples network definition, docstring, and examples 3 validation loop

Fabio-83:
## Related issues

example 3 update based on #215 

## Description

Updated the network definition in the examples and related docstring.
Validation function on example 3 now validate over the entire dataset rather than the last image of the batch.


diego-plan9 (approved):
Thanks Fabio!

PR #238, opened 2021-06-02
Update to epoch loss calculation

Fabio-83:
## Related issues

Loss calculation in some of the examples was higher for the validation set than the training set

## Description

fix in the validation loss calculation and update to the hook call for the training and validation runner metrics


diego-plan9 (approved):
Thanks @Fabio-83 !

Issue #48, opened 2020-10-27
Add new optimizer: Adam


diego-plan9:
## Description and motivation

As part of widening the scope of the library, it would be great to include more analog optimizers that can be used in different context and use cases. One of the likely first candidates would be one of the Adam variations due its popularity.

## Proposed solution

In a similar way to `AnalogSGD`, create a new subclass of the chosen `torch` optimizer, customizing its behavior when passing through the analog layers. 

## Alternatives and other information

* As this would be the second optimizer, some refactoring is also a likely candidate in order to generalize the common functionality for both optimizers (thank Corey for the suggestion!) - in particular, `regroup_param_groups()` seems like a good candidate for a generic function to be shared by all analog optimizers through a base class or similar means.
* Longer term, we should combine this approach with others that allow us to "adapt" other non-analog optimizers more easily - for example, making the tiles behave more like tensors or dividing the optimizers between analog and non-analog. These approaches are a bit more far-fetched

diego-plan9:
Closing via #242.

In that PR, a generic analog optimizer is introduced that can be used with any regular optimizer - passing it as the first argument of the constructor, followed by the regular arguments for the desired optimizer. In the case of `Adam`, for example:

```python
from torch.optim import Adam
from aihwkit.optim import AnalogOptimizer

my_optimizer = AnalogOptimizer(Adam, ...)
```

The PR also includes the tooling for creating a class directly if desired, which might be more convenient in some cases:
```python
from torch.optim import Optimizer, Adam
from aihwkit.optim.analog_optimizer import AnalogOptimizerMixin

class AnalogSGD(AnalogOptimizerMixin, Adam):
    pass
```

PR #241, opened 2021-06-04
Hardware-ware training with weight gradients, AnalogContext

### LINKED ISSUE #94: Enhance decay_weights in AnalogSGD
## Description and motivation

Currently there are some aspects of using `decay_weights` in `AnalogSGD` that are a bit different from the standard `SGD`. It would be great if we could refine the remaining aspects and revise the current documentation in order to highlight the differences and potential pitfalls to users, and streamline a bit its usage.

This is meant to be a small, "refinement" issue - not meaning to include bigger changes to the optimizer.

### LINKED ISSUE #211: Cleaner way of passing values from backward to update
## Description and motivation

Currently, we are passing the values needed for the `update()` by "attaching" them as attributes of the `layer.weights` tensor (during `backward()`). We should try to find a better way, perhaps using `model.buffers` instead.

maljoras:
## Related issues

makes foundation of #48 
closes #94  (any torch weight decay can now be used for inference tiles for hardware aware training)
closes #211 

The use of `weight` and `bias` as the proxy for the analog updater was awkward, since these were added as parameters and then removed again in `AnalogSgd`.  Especially for inference one might want to use the standard torch optimizers.  

## Description
Now inference tiles uses shared weights, which means that all standard torch optimizers can be used in this case in principle. This is because, the inference tiles now return the weight gradient in the backward pass. For analog training, the weight gradient is still directly applied in-memory as before.    

## Details

* A new `AnalogContext` is introduced which inherits from `Parameter`. This is the only proxy needed for the `AnalogSGD`. 
* In case of "shared_weights" (for inference only) weights are set to the torch memory and handled as a normal Parameter. Thus all SGD should work for inference tiles (including distributed training in principle).
* Regrouping is not necessary any more, because only an `AnalogContext` is inserted as parameter, and not the weights/bias of AnalogLinear etc. layers.
* weight / bias field is still used for syncing as before, but `state_dict` will not save these values, since they are not `Parameters` anymore. The weight is handled by the tile instead.   
* Also fixes a segfault in case of shared weights usage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,6 @@ The format is based on [Keep a Changelog], and this project adheres to
 
 ### Added
 
-* A new `AnalogLSTM` module: a recurrent neural network that uses
-  AnalogLinear. (\#240)
 * A number of new config presets added to the library, namely `EcRamMOPreset`,
   `EcRamMO2Preset`, `EcRamMO4Preset`, `TikiTakaEcRamMOPreset`,
   `MixedPrecisionEcRamMOPreset`. These can be used for tile configuration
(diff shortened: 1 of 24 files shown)


diego-plan9 (approved):
Great addition Malte! So glad to finally be able to close #211 and for the new features overall.

PR #243, opened 2021-06-04
PCM Preset refresh once per mbatch (OneSidedUnitCell)  

maljoras:
## Related issues

CPU and GPU did different refresh cycles by default for `OneSidedUnitCell` ( on which `PCMPreset` is based on).  

## Description

* This adds a `units_in_mbatch` switch to the refresh for `OneSidedUnitCell` and sets it by default on once per `mbatch` (ie the array reuse factor for one mini-batch). 
* Note that for the GPU version, refresh can only be done once per `mbatch`, for CPU can do also once per mat-vec.


diego-plan9 (approved):

diego-plan9:
Merging as the travis failure will be fixed at #244

PR #244, opened 2021-06-04
Fix occasional unit test error due to noisy tile

maljoras:
## Related issues

The unit test resulted in occasional failures because of the noise in the `ConstantStep` tile, which is irrelevant for the unit test.  

## Description

Added a new parameterized tile class that removes the noise parameters to form a `IdealizedConstantStep` that can be used iin tests when appropriate.


diego-plan9 (approved):
Thanks for the quick fix!

PR #250, opened 2021-06-14
CUB namespace

maljoras:
## Related issues

CUB headers use static variables which results in CUDA errors for some configurations when using it together with pytorch's CUB and Thrust headers (known issue in pytorch).   

## Description

Use a private namespace for all CUB calls in RPUCUDA.


diego-plan9 (approved):
Thanks Malte! For reference, this is related to https://github.com/pytorch/pytorch/issues/52663.

PR #252, opened 2021-06-14
Move to cuda using analog context

maljoras:
## Related issues

`model.cuda()` would not move tiles to CUDA if analog tiles reside inside a module that is derived from `nn.Module`.    

## Description

Now the tile cuda move is handled in the `AnalogContext`


diego-plan9 (approved):
:tada:

PR #257, opened 2021-06-17
Removed CudaXTiles

### LINKED ISSUE #253: Question about save / load model
## Description

 I have saved the model after training, but it will get the error below when loading the model and using to inference

RuntimeError: x_input must be a CPU tensor

## How to reproduce

Save the model -> load the model and run either inference or training

## Expected behavior

It can load model and inference or retraining fluently

## Other information

* Pytorch version: 1.8.1+cu102
* Package version: 
* OS: Ubuntu
* Python version: 3.6
* Conda version (or N/A): N/A

maljoras:
## Related issues

fixes #253 

## Description

The move of an analog tile by using `cuda()` to GPU involved construction of a C++ CUDA tile (in the `.tile` field) as well as changing the python tile class from e.g. `AnalogTile` to `CudaAnalogTile`. The python class change was cumbersome, as it did not comply with the in-place form of moving a tile to cuda usually used in pytorch. Now only the CUDA C++ tile class will be constructed within the `AnalogTile` (so the `tile` field does either hold a C++ `rpu_base.AnalogTile` or a `rpu_base.CudaAnalogTile`) thus behaving more like torch's `Tensor` which can be Cuda or Cpu. The python `CudaAnalogTile` class (and similarly the e.g. `CudaInferenceTile` class) are now deprecated.      

This solves a number of issues of moving tiles to cuda.


maljoras:
Thanks to @diego-plan9 for adding more tests!

diego-plan9 (approved):
Thanks Malte!

PR #261, opened 2021-06-21
Fix aihwkit.simulator.tiles imports

diego-plan9:
## Description

Since a few commits ago, some of the imports (namely `from aihwkit.simulator.tiles import Foo`) were not working if executed directly due to circular imports. This PR revises the `aihwkit.simulator.tiles` to ensure they avoid importing items from `aihwkit.simulator.configs` continuing the effort in  #213 , by ensuring that the conflicting elements are imported dynamically.


diego-plan9:
Merging despite CI not passing - it seems one of the jobs is taking longer than usual in queue.

PR #281, opened 2021-08-03
Reorganization of the module structure for inference

maljoras:
## Related issues

* The module `noise_models` was packed with different classes, which made extensions difficult.  
* The statistical noise models are only applicable for inference and thus a separate module should exist
* These inference related noise models are strictly not part of the core simulator, since weight matrices are modified externally, thus they should reside outside of the `aihwkit.simulator` tree   

## Description
To pave the way for adding new noise models for inference, the different classes are now in separated python files under the directory `aihwkit.inference`. Thus, the import statement for noise models for inference now changes to `from aihwkit.inference` eg.:
```python
from aihwkit.inference import PCMLikeNoiseModel
```
The earlier imports such as `from aihwkit.simulator.noise_models import PCMLikeNoiseModel` are now depreciated and will result in a future warning. It will be removed soon.  

## Details
* Also pylint/pycodestyle is now enabled for the examples and tests, which made a number of small fixes necessary (mostly to the documentation)


diego-plan9 (commented):
Nice reorganization, @maljoras ! And a well-needed one - this should indeed ease the way for adding new noise models (and related entities) to the library in a neatly structured way, and also makes them easier to consume by users.

Thanks as well for including the changes needed for the API documentation, they look flawless - and also for the boost in readability to the examples. :clap: 

Actually I only have one meaningful comment: could you add an `__init__.py` file to the three new packages under `aihwkit.inference`, in order to make them "real" packages and prevent some tools chocking on them in the future?

maljoras:
Many thanks, @diego-plan9!
I have added  the init-files as you suggested.

diego-plan9 (approved):

Issue #306, opened 2021-11-09
Bias term left as pytorch nn.Parameter


Fabio-83:
## Description and motivation

At present, when the bias term is requested from the layers, an additional column in the `rpu_tile` is created and the biases are updated during the analog backward pass, which add noise and non-idealities typical of the analog devices. In some hardware design it may be desirable to compute the bias in digital, as it would improve the overall network accuracy.

## Proposed solution

Now the bias can be either compute in the analog `rpu_tile` or left as `pytorch` `nn.Parameter`. This behavior can be controlled by setting the `digital_bias` parameter of the layers to `True`.

maljoras:
Addressed in #307

PR #307, opened 2021-11-09
Digital bias

Fabio-83:
## Related issues

#306 

## Description

Until now, when the bias term is requested from the layers, an additional column in the rpu_tile is created and the biases are updated during the analog backward pass, which add noise and non-idealities typical of the analog devices. In some hardware design it may be desirable to compute the bias in digital, as it would improve the overall network accuracy.

## Details

Now the bias can be either compute in the analog rpu_tile or left as pytorch nn.Parameter. This behavior can be controlled by setting the digital_bias parameter of the layers to True.


maljoras (approved):
Great, many thanks @Fabio-83

PR #312, opened 2021-11-09
More non-idealities of analog mat-vec & transfer compound enhancements

### LINKED ISSUE #308: More realistic analog non-idealities: IR-drop
## Description and motivation

Currently, IR-drop is not simulated during the analog-mat vec  

## Proposed solution

Implement  IR-drop for CPU and GPU.

### LINKED ISSUE #309: Transposed transfer for TransferCompound
## Description and motivation

Currently only a forward read is supported for the Tiki-taka learning rule, when transferring from the fast array to the slow array. 

## Proposed solution

Add an option to use a backward read to transfer.  This might improve learning for rectangular shaped weight matrices.

### LINKED ISSUE #310: State-dependent short-term PCM like read noise
## Description and motivation

PCM shows a state-dependency in the read noise, this is currently not modelled. 

## Proposed solution

Add a new weight noise type, that reversibly adds output noise that is proportional to the flowing current in a way reasonable for PCM devices.

maljoras:
## Related issues

closes #309,  closes #308, closes #310 

## Description

Update of the C++ backend enabling more analog non-idealities and new features in the `TransferCompound` 
Note that some parameter names have changed for `TransferCompound` (those referring to `columns` due to now supporting also rows) 

## Details
New features include: 
* Transposed read for transfer device (`transfer_columns`)
* PCM weight noise 
* IR drop along columns 
* enable dense updates with more devices

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,10 @@ The format is based on [Keep a Changelog], and this project adheres to
 * State indipendent inference noise model. (\# 284)
 * Transfer LR parameter for ``MixedPrecisionCompound``. (\#283)
 * The bias term can now be handled either by the analog or digital domain by controlling
-  the `digital_bias` layer parameter. (\#306 to be updated)
+  the `digital_bias` layer parameter. (\#307)
+* PCM short-term weight noise. (#???)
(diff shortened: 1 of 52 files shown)


Fabio-83 (approved):
Thanks @maljoras

Issue #302, opened 2021-11-05
Question related to #106 (Weight Initialization)


jiwonjae:
Hi,

I have questions related to #106 .
It was answered that initialization of weights in `TransferCompound` can be done by "set_hidden_parameters", yet I couldn't change the weight values of each array inside the unit cell.

Here is the code I used:

**1) Device used**

```python
device=TransferCompound(
# Devices that compose the Tiki-taka compound.
unit_cell_devices=[
LinearStepDevice(),
LinearStepDevice()
],
# Make some adjustments of the way Tiki-Taka is performed.
units_in_mbatch=True, # batch_size=1 anyway
transfer_every=1, # every 2 batches do a transfer-read
n_cols_per_transfer=1, # one forward read for each transfer
gamma=1,
scale_transfer_lr=False, # in relative terms to SGD LR
transfer_lr=0.01, # same transfer LR as for SGD
)
rpu = UnitCellRPUConfig(device=device)
```

**2) setting up hidden weights - gamma = 1**

```python
analog_tile = AnalogTile(out_size=1,
in_size=1,
rpu_config=rpu)



temp = analog_tile.get_hidden_parameters()
print("original value for weight1: ",temp["hidden_weights_0"])
print("original value for weight2: ",temp["hidden_weights_1"])
print("original value for max bound1: ", temp["max_bound_0"])
print("original value for dwmin_up2: ", temp["dwmin_up_1"])
print(" ")
temp["hidden_weights_0"] = torch.tensor(1.0).repeat((1,1)).transpose(0,1)
temp["hidden_weights_1"] = torch.tensor(1.0).repeat((1,1)).transpose(0,1)
temp["max_bound_0"] = torch.tensor(123.0).repeat((1,1)).transpose(0,1)
temp["dwmin_up_1"] = torch.tensor(987.0).repeat((1,1)).transpose(0,1)
print("setting for weights saved in 'temp' :",temp["hidden_weights_1"])
print("setting for max bound1 saved in 'temp' :",temp["max_bound_0"])
print("setting for dwmin_up2 saved in 'temp' :",temp["dwmin_up_1"])
print(" ")
print("'temp'")
print(temp)



analog_tile.set_hidden_parameters(temp)
print(" ")
print("initial setting done for weight1: ",analog_tile.get_hidden_parameters()["hidden_weights_0"])
print("initial setting done for weight2: ", analog_tile.get_hidden_parameters()["hidden_weights_1"])
[...]

PR #313, opened 2021-11-09
Fixed hierachical setting of hidden weights

### LINKED ISSUE #303: Set/get slow weight of a transfer compound through hidden parameters in case gamma equal zero
## Description and motivation

As mentioned in #302, it is a little counter-intuitive that for `gamma=0` slow weights cannot be accessed by `set_hidden_parameters`.   

## Proposed solution

Make sure that setting hidden parameters sets all hidden weights as expected even for `gamma=0`

maljoras:
## Related issues

closes #303,  Hierarchical hidden parameter settings would not update the weight properly.  
related to #302 

## Description

The hidden weights now update the weights. Note that for multiple hidden weights hierachy the lowest will determine the upper (higher) levels.  

## Details

Now the explicit setting of the hidden weights is possible (see below). 

Note, however, that in case of `gamma=0`, the reading of the slow matrix might be impacted (shows zeros when get_hidden_parameters). In this case, the matrix needs to be read with `getWeights()`.    

Note also that the setting of the weights with lowest level (e.g. `hidden_weights_0_1`) will determine (and overwrite) the weight on the next level (e.g. `hidden_weight_1`).

Here an example:

```python


from torch import ones
from aihwkit.simulator.configs import UnitCellRPUConfig
from aihwkit.simulator.configs.devices import (
    TransferCompound,
    ReferenceUnitCell,
    SoftBoundsDevice)


# The Tiki-taka learning rule can be implemented using the transfer device.
def SBdevice():
    """Custom device """
    return SoftBoundsDevice(w_max_dtod=0.0, w_min_dtod=0.0, w_max=1.0, w_min=-1.0)

# arbitrary
gamma = 0.2

rpu_config = UnitCellRPUConfig(
    device=TransferCompound(

        # Devices that compose the Tiki-taka compound.
        unit_cell_devices=[
            ReferenceUnitCell([SBdevice(), SBdevice()]),  # fast "A" matrix
            ReferenceUnitCell([SBdevice(), SBdevice()])   # slow "C" matrix
        ],
        gamma=gamma,
    )
)

# print the config
print(rpu_config)

# construct the model
in_features = 4
out_features = 2
model = AnalogLinear(in_features, out_features, bias=True, rpu_config=rpu_config)

# set the reference devices to some values
params = model.analog_tile.get_hidden_parameters()
shape = params['hidden_weights_0_0'].shape

# just dummy settings
a, b, c, d = 0.4, 0.2, 0.6, 0.1
params['hidden_weights_0_0'] = a*ones(*shape)  # A
params['hidden_weights_1_0'] = b*ones(*shape)  # A ref
params['hidden_weights_0_1'] = c*ones(*shape)  # C
params['hidden_weights_1_1'] = d*ones(*shape)  # C_ref

model.analog_tile.set_hidden_parameters(params)

weight, bias = model.get_weights()

# should be true for all weights
print(weight == gamma*(a - b) + c - d)

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog], and this project adheres to
 * Map location keyword for load / save observed. (\#293)
 * Fixed issue with CUDA buffer allocation when batch size changed. (\#294)
 * Fixed missing load statedict for AnalogSequential. (\#295) 
+* Fixed issue with hierachical hidden parameter settings. (\#313)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ The format is based on [Keep a Changelog], and this project adheres to
 * Map location keyword for load / save observed. (\#293)
 * Fixed issue with CUDA buffer allocation when batch size changed. (\#294)
 * Fixed missing load statedict for AnalogSequential. (\#295) 
-* Fixed issue with hierachical hidden parameter settings. (\#313)
+* Fixed issue with hierarchical hidden parameter settings. (\#313)
 
 ### Changed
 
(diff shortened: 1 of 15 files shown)


Fabio-83 (approved):
Thanks a lot @maljoras

Issue #285, opened 2021-09-14
Analog model information


maljoras:
## Description and motivation

For performance evaluation it would be great to have a tool similar or based on `torchinfo` that shows  for example (but no limited to)
* the number of parameters on digital and analog, 
* how many analog tiles are used in the model (and their sizes), 
* how many digital mult-adds are used 
* how analog tile operations (analog mat-vecs, analog outer product) 

a given model has. Note that the analog optimizers often have a mixture of digital and analog operations which thus should be counted accordingly.  

## Proposed solution

For each analog model with a given  `rpu_config` the information needs to be calculated. Note that based on the `rpu_config` type, number of operations are very different, thus it might make sense to add info methods to the config classes.  

Ideally there should be a way to return the info not only in a print out table form as below, but also as a e.g. dictionary for connecting to other tools on performance evaluation.      

For instance, output could be similar to `torchinfo` but with additional columns with information about analog components:

```python
from torchinfo import summary
summary(model, input_size=(1, 3, 224, 224), col_names=['input_size', 'output_size',  'kernel_size', 'mult_adds'])
```
```
============================================================================================================================================
Layer (type:depth-idx)                   Input Shape               Output Shape              Kernel Shape              Mult-Adds
============================================================================================================================================
ResNet                                   --                        --                        --                        --
├─Conv2d: 1-1                            [1, 3, 224, 224]          [1, 64, 112, 112]         [3, 64, 7, 7]             118,013,952
├─BatchNorm2d: 1-2                       [1, 64, 112, 112]         [1, 64, 112, 112]         [64]                      128
├─ReLU: 1-3                              [1, 64, 112, 112]         [1, 64, 112, 112]         --                        --
├─MaxPool2d: 1-4                         [1, 64, 112, 112]         [1, 64, 56, 56]           --                        --
├─Sequential: 1-5                        [1, 64, 56, 56]           [1, 64, 56, 56]           --                        --
│    └─BasicBlock: 2-1                   [1, 64, 56, 56]           [1, 64, 56, 56]           --                        --
│    │    └─Conv2d: 3-1                  [1, 64, 56, 56]           [1, 64, 56, 56]           [64, 64, 3, 3]            115,605,504
│    │    └─BatchNorm2d: 3-2             [1, 64, 56, 56]           [1, 64, 56, 56]           [64]                      128
│    │    └─ReLU: 3-3                    [1, 64, 56, 56]           [1, 64, 56, 56]           --                        --
[...]

PR #316, opened 2021-11-12
Analog info

IHIaadj:
## Related issues

"Analog model information" Enhancement #285. 

## Description
Structure of the Analog info modules that provides a summary information of the model executed on the RPU configuration. 

## Details
The structure is composed of two classes: 
- **LayerInfo**: stores the information at the layer level.  
            - Name 
            - IsAnalog 
            - Input Shape 
            - Output Shape 
            - Kernel Size 
            - # of Tiles 
            - # of Multi-Adds ops 
            - Reuse factor 
- AnalogInfo: stores and computes the different statistics. 
            - contains a list of all LayerInfo of the model 
            - computes the total number of tiles and total number of parameters. 

A small example on Lenet is added in the /examples folder. 

Need to be done: 
        - fix the number of parameters according to the used RPU configuration.


src/aihwkit/utils/analog_info.py
```diff
@@ -0,0 +1,178 @@
...
+    def calculate_size(self, input_size):
+        """Set input_size or output_size of the layer."""
+        rand_tensor = torch.rand(input_size)
+        output_size = self.module(rand_tensor).shape
+        self.__set_layer_size(input_size, tuple(output_size))
+        return input_size, tuple(output_size)
+
+    def set_kernel_size(self): 
+        try: 
+            self.kernel_size = self.module.kernel_size
```
maljoras, line 70: please use `hasattr` rather than try/catch

maljoras (changes requested):
@IHIaadj excellent ! Very nice overall. I haven't looked into into detail right now (I will do so soon) but until now, could you make sure that all the formatting requirements are met ? That means, the commands: 
``` make mypy; make pylint; make pycodestyle; make pytest ```
all should have no error message.

File: src/aihwkit/utils/analog_info.py [hunk 2]
```diff
@@ -0,0 +1,203 @@
...
+        """Create the layer summary list.
+
+        This list contains LayerInfo elements that corresponds to each layer of the model."""
+        layer_summary = []
+        if isinstance(self.model, tuple):
+            layer_summary.append(LayerInfo(self.model[0], self.rpu_config))
+        else:
+            for module in self.model.modules():
+                if not isinstance(module, nn.Sequential):
+                    layer_summary.append(LayerInfo(module, self.rpu_config))
```
maljoras, line 148: Each tile has its own `rpu_config` in principle. So best would be if the rpu_config for a tile is taken here. Shortly, I will update the master branch with a new commit that can loop over analog_tiles() for an AnalogModuleBase. One could use that one.

src/aihwkit/utils/analog_info.py
```diff
@@ -0,0 +1,203 @@
...
+        The reuse factor is the number of vector matrix multiplication a layer computes."""
+        if isinstance(self.module, AnalogLinear):
+            ruf = reduce(operator.mul, (self.input_size), 1) // int(self.input_size[-1])
+            self.__set_reuse_factor(ruf)
+        elif isinstance(self.module, _AnalogConvNd):
+            # TODO: Extend the formula to ConvNd
+            ruf = reduce(operator.mul, (self.output_size), 1) // self.output_size[1]
+            self.__set_reuse_factor(ruf)
+
+    def layer_summary_dict(self) -> dict:
```
maljoras, line 95: Can one add the (logical) tile size, ie the weight matrix size stored on the tile?  Also utilization might be good to add later, once we map to multiple tiles (will be possible soon)

src/aihwkit/utils/analog_info.py
```diff
@@ -0,0 +1,203 @@
...
+
+def analog_summary(model: nn.Module,
+                   input_size: Optional[INPUT_SIZE_TYPE] = None,
+                   rpu_config: Optional[RPUConfigAlias] = None) -> AnalogInfo:
+    """
+    Summarize the given PyTorch model. Summarized information includes:
+        1) Layer names,
+        2) input/output shapes,
+        3) kernel shape,
+        4) # of digital parameters,
```
maljoras, line 184: This one might depend on the rpu_config settings.

src/aihwkit/utils/analog_info.py
```diff
@@ -0,0 +1,203 @@
...
+from aihwkit.nn import AnalogLinear
+
+COLUMN_NAMES = {
+    "name": "Layer Name",
+    "isanalog": "Analog/Digital",
+    "input_size": "Input Shape",
+    "output_size": "Output Shape",
+    "kernel_size": "Kernel Shape",
+    "num_tiles": "Number of Tiles",
+    "macs": "Multi_Adds",
```
maljoras, line 26: Are the macs computed?

maljoras:
@IHIaadj , we have now added the `AnalogLinearMapped` module. I wonder whether you can added that for the info? you can use something like  
```python
for tile in module.analog_tile():
    # tile.in_size, tile.out_size to get the actual size (of the mapped weights)
    # tile.rpu_config.mapping.max_input_size / max_output_size to get the "physical" tile size. 
    # this can be used to compute utilization 
```

maljoras:
Hi @IHIaadj ,
I have merged the new layers which have multiple analog tiles. Could you update your code that it also works with those layers (AnalogLinearMapped, AnalogConv1,2,3Mapped)? 

Then we can merge your PR. Thanks a lot again!

IHIaadj:
Is the reuse factor shared with all the layers' tiles ?

maljoras:
@IHIaadj right. The re-use factor is shared for all the tiles of a mapped layer. It just has more tiles where the weight matrix is split across. One needs technically more digital compute because the output of all the tiles in the in-dimension are added up in digital. So one needs additional n_in_tiles*out_features where n_in_tiles are the number of tiles in the input dimension. For the output dimension one needs also additional n_out_tiles*in_features but only during backward pass.

maljoras:
@IHIaadj, are you finsihed with you and addition so that I can start a review? It seems that there are still a number of formatting (pylint, mypy,  pycodestyle) issues though.

IHIaadj:
@maljoras please review it now. I fixed the python style issues. And I created two categories in the table final information (Layer Info and Tile Info)

maljoras (changes requested):
Many thanks again. It looks great but it seems not working in some cases: 

1) When I add `analog_summary(model, (1, 3, 224, 224))` to example 17, it seems that the rand_input is of shape [1, 1000] which is not correct for convolutions. It throws an error. Can you check what is wrong? In case of a purely digital resnet34 it also says: 

```
RuntimeError: Expected 4-dimensional input for 4-dimensional weight [64, 3, 7, 7], but got 2-dimensional input of size [1, 1000] instead
```
 
2) For the example 19 the output looks like:
image
* For the tile information, why are 2 lines used (the first is - - - - )
* Why is the utilization 0 ? For non-mapped layers the utilization should be always 1.0 actually
* Can one have a summary line at the end how many analog MACs were performed (adding all the reuse factors) as well as the amount of digital MACs used (there might be digital computations still in the network).

maljoras:
Another possible addition could be to also print the class of the RPUConfig somewhere (for instance in isanalog), for instance "InferenceRPUConfig". But we can also add this later

IHIaadj:
Hello @maljoras, 

I'm fixing the first issue with resnet34. There's a skip connections that are not handled. I'll fix this. 

**For example 19:** 
> For the tile information, why are 2 lines used (the first is - - - - )

The table is structured as follows. For each layer, we display the information of the layer and the number of tiles (let's say n tiles) used and then the n next layer will be dedicated to the tile information. This is why the first line contains --- in the tile info section. 

> Why is the utilization 0 ? For non-mapped layers the utilization should be always 1.0 actually

This is what I supposed. But the actual physical size of the tile created is 512. Although I'm using non-mapped layers. 

> Can one have a summary line at the end how many analog MACs were performed (adding all the reuse factors) as well as the amount of digital MACs used (there might be digital computations still in the network).
In the torch's summary there is also a line on how many free parameters. Would be good if we could retain that, but distinguishing analog versus digital parameters.

I'll add that. The number of Analog parameters depends on the type of device used right ? Is it computed somewhere or I just use the number of used capacity in each tile ? 

> Why is there a separate "Per-layer information" (with no info) and a "layer information" section?
Actually, I wanted two section the "Per-layer information" (where we have two categories: layer and tile information) and "General Model information" where we can have the MACs, total number of tiles, size. 

> The utilization is in % ? Seems that there is missing a factor of 100 then ?
True. there's a *100 missing. Fixed.

maljoras:
@IHIaadj For the utilization computation, it is a little confusing in the AIHWKIT code right now. To make it most consistent, best would be if you would only compute the utilization in case of "Mapped" layers. Otherwise just print a "N/A".

src/aihwkit/utils/analog_info.py
```diff
@@ -0,0 +1,287 @@
...
+COLUMN_NAMES = {
+    "name": (0, "Layer Name"),
+    "isanalog": (0, "Is Analog"),
+    "input_size": (0, "In Shape"),
+    "output_size": (0, "Out Shape"),
+    "kernel_size": (0, "Kernel Shape"),
+    "num_tiles": (0, "# of Tiles"),
+    # "macs": "Multi_Adds",
+    "log_in_size": (1, "in_size (log)"),
+    "log_out_size": (1, "out_size (log)"),
```
maljoras, line 33: I would prefer if one would use a single `log_shape` instead of two columns for in / out size:   `Log. tile shape (out, in)` could be good title and then display the shape `(out, in)`. 
Similarly one could add a column with 'Phys. tile shape (out, in)`

src/aihwkit/utils/analog_info.py
```diff
@@ -0,0 +1,287 @@
...
+        analog = "analog" if self.isanalog else "digital"
+        return {"name": self.name,
+                "isanalog": analog,
+                "input_size": str(self.input_size) if self.input_size is not None else "-",
+                "output_size": str(self.output_size) if self.output_size is not None else "-",
+                "kernel_size": str(self.kernel_size) if self.kernel_size is not None else "-",
+                "num_tiles": self.num_tiles,
+                "reuse_factor": str(self.reuse_factor) if self.reuse_factor is not None else "-",
+                "log_in_size": "-",
+                "log_out_size": "-",
```
maljoras, line 159: phys_in/out might be useful here to add

src/aihwkit/utils/analog_info.py
```diff
@@ -0,0 +1,287 @@
...
+        result += "General Information\n" + divider
+        result += "Total number of tiles: " + str(self.total_tile_number) + "\n"
+        result += "Total number of analog layers: " + str(self.total_nb_analog) + "\n"
+        return result
+
+
+def analog_summary(model: nn.Module,
+                   input_size: Optional[INPUT_SIZE_TYPE] = None,
+                   rpu_config: Optional[RPUConfigAlias] = None) -> AnalogInfo:
+    """
```
maljoras, line 263: Is there an easy way for the use to select the columns displayed? like torchinfos "col_names". https://pypi.org/project/torchinfo/
But we can also tackle this as a follow-up

src/aihwkit/utils/analog_info.py
```diff
@@ -0,0 +1,287 @@
...
+            self.kernel_size = self.module.kernel_size
+
+    def calculate_reuse_factor(self) -> None:
+        """Compute the reuse factor.
+        The reuse factor is the number of vector matrix multiplication a layer computes."""
+        if isinstance(self.module, AnalogLinear):
+            ruf = reduce(operator.mul, (self.input_size), 1) // int(self.input_size[-1])
+            self.__set_reuse_factor(ruf)
+        elif isinstance(self.module, (_AnalogConvNd, _AnalogConvNdMapped)):
+            # TODO: Extend the formula to ConvNd
```
maljoras, line 136: is this still a TODO?

src/aihwkit/utils/analog_info.py
```diff
@@ -0,0 +1,287 @@
...
+            result += (("{:<20}"*len(header_i)).format(*header_i))
+        result += "\n"
+
+        for x in self.layer_summary:
+            result += repr(x)
+
+        result += divider
+        result += "General Information\n" + divider
+        result += "Total number of tiles: " + str(self.total_tile_number) + "\n"
+        result += "Total number of analog layers: " + str(self.total_nb_analog) + "\n"
```
maljoras, line 256: We might want to add more summary information, like digital-MACs runtime etc later. We can add this to a follow-up issue

maljoras (dismissed):
Great addition, many thanks @IHIaadj !
There are a few things we can improve and extend going forward, but we can do that in a follow-up issue as well. For instance:
1) combine in/out size into one column
2) add phys tile shape column
3) add some more summary info like remaining digital-MACs, total number of analog tile mat-vecs 
4) user-definable "columns" to select only some of the columns to display
5) maybe somehow improve the (- - - ) first entry of the tile section. It is a little confusing

src/aihwkit/utils/analog_info.py
IHIaadj: What should I print for the physical shape when the layer is not mapped ?

maljoras: Just the logical size

IHIaadj: Just fixed it and added Phys. tile shape and Log. tile shape.

maljoras: great!

maljoras:
@IHIaadj 
Regarding the examples 19, can you also add a brief summary to the readme https://github.com/IBM/aihwkit/blob/master/examples/README.md ? Thanks!

maljoras:
Finally, please add also a one-liner to the CHANGELOG (\#316). I think then we can finally merge it! For the rest of the items we simply open a new follow-up issue to be tackled later.

maljoras:
@kaoutar55 we might also want to add a short description of the analog info into the documentation somewhere?

maljoras (changes requested):
@IHIaadj 
I actually found a small issue still: The re-use factor for a linear layer should be 1. it seems to display 0 ?

```
AnalogLinearMapped  analog              [1, 512]            [1, 1000]           -                   2                   0                   -                   -                   -                   
                                                                                                                                            (500, 512)          (512, 512)          97.66               
                                                                                                                                            (500, 512)          (512, 512)          97.66  
```

IHIaadj:
> @IHIaadj I actually found a small issue still: The re-use factor for a linear layer should be 1. it seems to display 0 ?
> 
> ```
> AnalogLinearMapped  analog              [1, 512]            [1, 1000]           -                   2                   0                   -                   -                   -                   
>                                                                                                                                             (500, 512)          (512, 512)          97.66               
>                                                                                                                                             (500, 512)          (512, 512)          97.66  
> ```

Oups sorry, fixed.

maljoras (approved):
Great addition, many thanks @IHIaadj

PR #319, opened 2021-11-12
Decay to reset bias value 

### LINKED ISSUE #315: Decay weight value to arbitrary point
## Description and motivation

Currently, the `lifetime` parameter decays the weight value exponentially to 0. However, for some materials the decay might not go to strictly zero.  

## Proposed solution

Add a "decay_to" parameter that sets the decay to value for each device.

maljoras:
## Related issues

closes #315 

## Description

Decay (set with lifetime) is now changed for pulsed devices  to decay to the `reset_bias` which is given by the `reset` (mean) and `reset_dtod` parameters. 

## Details

Please take a look at this [test](https://github.com/IBM/aihwkit/blob/master/tests/test_specific_tiles.py#L96) to see how one could set arbitrary decay points by setting the `reset_bias`.

### DIFF
--- a/src/aihwkit/simulator/configs/devices.py
+++ b/src/aihwkit/simulator/configs/devices.py
@@ -86,7 +86,7 @@ class PulsedDevice(_PrintableMixin):
 
     **Decay**:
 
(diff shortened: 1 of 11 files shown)


Fabio-83 (approved):
thanks @maljoras

Issue #314, opened 2021-11-10
Questions about using tiles


ElinorG11:
Hello,

I have a few questions about using the tiles:

1. what is exactly the difference between analog tile and inference tile? I thought that analog tile is an abstract object and we can't really use it, but as I seen in some of your examples you use it explicitly (as seen in example 01 where you use `SingleRPUConfig()` which uses `AnalogTiles`).

2. How can I specify the size of the tile after defining an rpu configuration? from looking at you documentation, I saw that when creating a tile, the user can specify it's size, as following `tile = InferenceTile(32, 32)`. but when creating a model or when using an existing one and converting it to analog counterpart, we use `rpu_config = InferenceRPUConfig()`. I'd appreciate if you could explain to me:
- What is the difference between the configuration and the tile itself? the configuration just allows the user to specify more parameters for a specific tile right?
- When defining configuration as follows `rpu_config = InferenceRPUConfig()` Suppose now I want to change the tile to be 32x32. how can I do it? by viewing you documentation I saw the functions `rpu_config.forward.inp_res` and `rpu_config.forward.out_res` which quantise the ADC and DAC. so should I specify `rpu_config.forward.inp_res = 1/32` and `rpu_config.forward.out_res = 1/32` or did I misinterpreted the meaning of the last two functions? I saw this example also:
```
from aihwkit.simulator.tiles import AnalogTile
from aihwkit.simulator.configs import SingleRPUConfig
from aihwkit.simulator.configs.devices import ConstantStepDevice

config = SingleRPUConfig(device=ConstantStepDevice())
tile = AnalogTile(32, 32, rpu_config=config)
```

how can I make sure this tile will be used in the configuration which is sent to the model as seen in example 03:

```
def create_analog_network(input_size, hidden_sizes, output_size):
   ...
    model = AnalogSequential(
        AnalogLinear(input_size, hidden_sizes[0], True,
                     rpu_config=SingleRPUConfig(device=ConstantStepDevice())),
        ...
```
- When using `InferenceRPUConfig` I saw that the default Tile type is InferenceTile but of what size (out_size and in_size)? and I also saw that you're using `ConstantStepDevice` as defult, but of what precision?

3. Is there any way to create our own configuration? this is, specify explicitly the tile with it's parameters, and the device and create an rpu configuration from it to send to the model?

Thank you very much for your help.

maljoras:
Hi @ElinorG11 ,
thanks for your questions !
[...]

PR #320, opened 2021-11-15
Linear layer mapped to multiple tiles of given physical size

maljoras:
## Related issues

related to #314 

## Description
This introduced a new layer module, `AnalogLinearMapped`, implements again an `Linear` like `AnalogLinear`. However, instead of `AnalogLinear` it will use multiple analog tiles of a given max physical size to implement linear layers where the weight matrix exceeds the physical size of a tile.   

## Details

* `rpu_config.mapping` with `MappingParameter()` let's the user set the physical tile size (`max_input_size`, `max_output_size`). Weight matrices will be split in a way that tiles are almost equally used. No additional noise comes from not programmed/used  columns/rows.


Fabio-83 (approved):
Thanks a lot @maljoras

PR #353, opened 2022-01-27
Adding learnable output scaling alpha

Fabio-83:
## Description

Moved the `weight_scaling_omega` parameter to the `MappingParameter` of the `rpu_config`.
Added a `weight_scaling_omega_columnwise` parameter which controls whether the scaling of the weights is performed using the global maximum or column-wise maximum.
Added a `learn_out_scaling_alpha` parameter to the `MappingParameter` of the `rpu_config` to control if the alpha scales can be learnable or fixed.

## Details

The `weight_scaling_omega` is now passed through the MappingParameter of the `rpu_config`. The `set_weight_scaling()` can be done either globally or column-wise. If global then the weights maximum is the matrix maximum, if column-wise then the maximum is taken for every column of the weights/biases matrix. The weight_max/weight_scaling_omega ratio is saved as `self.out_scaling_alpha` and is used to return the mathematical weights when `get_weights_scaled()`  is called.
If `learn_out_scaling_alpha` is set to `True`, then the `self.out_scaling_alpha` is registered as learnable parameter.


maljoras (approved):
Thanks you @Fabio-83 , great work!

PR #356, opened 2022-02-02
SelfDefine device based on user input

bobo-nums:
## Description

Implements a new device model class (```SelfDefineDevice```) where user can define up/down pulse and weights.


maljoras:
@bobo-nums @kaoutar55 I polished the device quite a bit. There were quite some unecessary loops in the device which I removed and optimized. You can take a look. I also enabled the full device-to-device variation. I also renamed the device and the fields to be more consistent with the other devices.  

Moreover I added a fitting tool to any device data. 

image

This is automatically fitted from: 
image

Maybe one could add fit to the variability as well (will add tomorrow).

Would be great of you would double check whether everything looks correct. 

I will try to add the CUDA part tomorrow.

maljoras:
@kaoutar55 I added the CUDA device and the write noise fitting in the example. Would be nice if one could add some documentation. The CSV read might also be added directly to the toolkit instead of to the example.

Black dots are the automatically fitted device data: 
image

maljoras (dismissed):
From my point of view we could merge it. Though before that we could also change/ add the following:
1. Add documentation of how to fit an experimental device
2. Add the device fit function from the example into the main code e.g. under aihwkit/utils/fitting.py
3. The fitting could be improved for odd bins, since the update step is not quite correct then (as the bin size is estimated for the middle of the bin)

kaoutar55:
Thank you @bobo-nums and @maljoras the new feature and update look great! Will work on adding documentation for this.

kaoutar55 (approved):

PR #358, opened 2022-03-09
Multiple RNN Cells and Layers

Jo7701:
Created a wrapper AnalogRNN class that accepts different cells (LSTM, GRU, VanillaRNN). Also has the option of creating a bidirectional layer. 

3 files were created: cells.py, layers.py, and rnn.py. All of these are in a rnn folder in nn/modules.


charlesmackin:
This looks great, thanks for the addition!

maljoras:
@Jo7701 many thanks for the contribution! If you could add the tests, then we would be able to merge it soon.

maljoras:
@Jo7701 Many thanks again for this great contribution. I took the liberty to polish your code somewhat.   It now works fine for CUDA as well. Thanks again, I will merge it very soon.

maljoras (approved):

kaoutar55:
Thank you @Jo7701 for these great additions! Thank you @maljoras for the detailed review and the enhancements.

Issue #346, opened 2022-01-10
Bump mypy version


maljoras:
## Description and motivation

Adjust the mypy version and unpin request, see #345.

PR #368, opened 2022-04-12
Drop Python 3.6, introduce Python 3.10

### LINKED ISSUE #364: Drop Python 3.6 support, and include 3.10
## Description and motivation

As Python 3.6 has reached end of life at the end of last year, it would make sense to drop support for it in `aihwkit` as well - and take the chance to introduce support for 3.10. In practice, this can be achieved by adjusting the CI so it takes `3.10` into account instead of `3.6` as the basis for checking the package compatibility with different versions (some tweaks derived from `3.10` might apply).

## Proposed solution

* update `travis.yml`
* update other references to `3.6` (mypy python version, `setup.py` minimal declared version)
* other tweaks needed from `3.10` support

N/A - https://devguide.python.org/devcycle/#end-of-life-branches

diego-plan9:
## Related issues

Closes #364

## Description

Make Python `3.7` the lowest supported version, along with checking for compatibility with Python `3.10` in CI, via:
* updating the travis jobs, with `3.7` as the "default" version and adding new `3.10` stages
* update the `cibuildwheel` variables in order to produce `3.10` wheels when a new version is released
* bump the minimal version in build metadata (and mypy syntax)

## Details

* There seems to be no warnings from `aihwkit` in 3.10 ([only some from `torchvision`](https://app.travis-ci.com/github/IBM/aihwkit/jobs/566778103#L3247))
* The bit in `visualization.py` is a quick fix for allowing `mypy` to pass ([errored](https://app.travis-ci.com/github/IBM/aihwkit/jobs/566770618#L723) in an earlier commit). A proper fix would be bumping `numpy` to a more recent version (some relation to #346 as well likely) - but it felt overkill to bump the version of one of the main libraries in this PR only due to typing.
* The `dataclasses` dependency has been dropped (as it was `<3.7` specific, and on 3.7+ is part of the standard library)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,7 +102,7 @@ jobs:
     env:
       - TORCH_VERSION_SPECIFIER="==1.7.1"
 
-  - name: "Build wheel for Python 3.6, 3.7, 3.8, 3.9 on Linux x86_64"
+  - name: "Build wheel for Python 3.7, 3.8, 3.9, 3.10 on Linux x86_64"
     os: linux
     dist: bionic
     language: python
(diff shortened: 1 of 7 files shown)


maljoras (approved):
Great, many thanks @diego-plan9 !

PR #373, opened 2022-05-11
GNU ABI autodetect and MKL libirc legacy dependence removed 

maljoras:
## Related issues

* GNU ABI needs to be the same as for pytorch. It is now detected from the torch installation
* In some situations the MKL library would add the IRC dependency which was a legacy in our findMKL module 

## Description

* removed libIRC dependency
* Auto-detect GNU ABI setting from zero


kaoutar55 (approved):
Many thanks @diego-plan9!

kaoutar55 (approved):

diego-plan9 (approved):
Looks good, thanks Malte!

PR #374, opened 2022-05-12
Added a fix in .travis.yml file for the last build error 

### LINKED ISSUE #371: How are the tests supposed to be run?
## Description and motivation

There are currently no main() functions for the tests, which makes it difficult for beginners.
Unlike the examples where everything is just straightforward.

## Proposed solution

Document the tests and how are they supposed to be used.

Make a main.py file for the tests folder.

kaoutar55:
Added a fix in .travis.yml file to the last build error that is caused by the error:

"The repository 'https://public.dhe.ibm.com/software/server/POWER/Linux/toolchain/at/ubuntu bionic Release' does not have a Release file."

The fix adds "sudo apt-get -qq install -y ca-certificates"
before any other apt-get commands.

Apparently some root certificate changed, so new CA certs are required to pull the release information. The error message is a bit misleading here.
the Release file is there, apt-get just can't retrieve it due to cert validation failures.

## Related issues

closes #371


maljoras (approved):
Thanks @kaoutar55 !

docs/source/api_reference.rst
```diff
@@ -28,7 +28,10 @@ API Reference
    aihwkit.nn.modules.container
    aihwkit.nn.modules.conv
    aihwkit.nn.modules.linear
-   aihwkit.nn.modules.lstm
+   aihwkit.nn.modules.rnn
+   aihwkit.nn.modules.rnn.cells
+   aihwkit.nn.modules.rnn.layers
+   aihwkit.nn.modules.rnn.rnn
```
diego-plan9, line 34: Non-blocking - originally, this file meant to contain only the Python _packages_ (ie. the directories) instead of all the modules (ie. the single `.py` files). The idea was to provide an overview rather than an exhaustive list of all modules, with the thinking that users could navigate to each package for more information about its modules (avoiding having a lot of cruft in this file, for example the `cloud` internals that are rarely used directly).

There are some exceptions to that policy already in this file in `master` (maybe due to some modules having particular importance) - but I wonder if it would make sense to just include `aihwkit.modules.rnn` in this case (and/or take a chance to unify, as for example `aihwkit.nn.modules.conv_mapped` and `linear_mapped` are missing, making it a bit inconsistent).

examples/README.md
```diff
@@ -1,7 +1,23 @@
 # IBM Analog Hardware Acceleration Kit: Examples
 
 We have many different examples to explore the many features of the IBM Analog Hardware
-Acceleration Kit:
+Acceleration Kit.
+
```
diego-plan9, line 5: Thanks for adding test-related information - can it be moved to a more sensible place, though? Usually, running the examples is a different use case than running the tests, which in my eyes is closer to contributing.

One of these places might be [`development_conventions.rst`](https://github.com/IBM/aihwkit/blob/master/docs/source/development_conventions.rst#testing) - there is a `Testing` section that can be expanded (and a mention to `pytest` earlier on). For more visibility, it would also make sense to add a link to that documentation page from the [contribution guidelines](https://github.com/IBM/aihwkit/blob/master/.github/CONTRIBUTING.md): that would keep the initial contribution lean, and still allow users to reach the full development conventions.

examples/README.md
```diff
@@ -1,7 +1,23 @@
...
+All the tests do not have a main.py file. We use Pytest for unit testing, which does not need any main.py file. A number of common developmental tools can be installed with pip install -r requirements-dev.txt
```
diego-plan9, line 6: Suggestion:

"This project contains a [unittest](https://docs.python.org/3/library/unittest.html) compatible test suite, that can be executed by any Python test runner. The recommended runner is `pytest`, which can be installed along with a number of other development tools used in `aihwkit` via:

```
pip install -r requirements-dev.txt 
```
"

The thinking is that the link would provide more material and information for users, including explanation about the structure and the lack of `main.py` - and we emphasize that the test can be run with other runners (but we favor `pytest`).

examples/README.md
```diff
@@ -1,7 +1,23 @@
...
+make pytest
+```
+
+To run individual test files, you can use, e.g.
+```
+pytest -v -s tests/test_presets.py
+```
+
+Individual tests can be run by giving the name, e.g.:
+pytest -v -s tests/test_presets.py::PresetTest_TTv2IdealizedPreset::test_tile_preset
```
diego-plan9, line 19: Can you also make this item a code block?

docs/requirements.txt
```diff
@@ -1,4 +1,5 @@
-Sphinx==3.1.2
+#Sphinx==3.1.2
```
diego-plan9, line 1: Can you remove this line?

diego-plan9 (commented):
Thanks Kaoutar! Added some small comments related to the (neat) documentation updates.

docs/source/api/aihwkit.nn.modules.rst
```diff
@@ -16,5 +16,6 @@ Submodules
    aihwkit.nn.modules.base
    aihwkit.nn.modules.container
    aihwkit.nn.modules.conv
+   aihwkit.nn.modules.conv_mapped
    aihwkit.nn.modules.linear
-   aihwkit.nn.modules.lstm
+   aihwkit.nn.modules.rnn
```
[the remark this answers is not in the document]
kaoutar55: good catch. I will add it also. Thank you!

examples/README.md
kaoutar55: Good point. I agree we should only show here how to run the examples rather than the unit tests. what do you think of moving this section to the redthedoc documentation? @diego-plan9

examples/README.md
kaoutar55: done. I also moved this to the tests directory. It is supposed to be added to the tests folder in a new Readme file

examples/README.md
diego-plan9: My suggestion would be:
* `examples/README.md` - keeping it the same as `master`, with information on how to run the examples.
* new information about running tests - add to `development_conventions.rst`, so it does end up in readthedocs [here](https://aihwkit.readthedocs.io/en/latest/development_conventions.html).
* additionally, add a link to [the rendered development conventions](https://aihwkit.readthedocs.io/en/latest/development_conventions.html) to `CONTRIBUTING.md`

The thinking is that running the tests is usually valuable for users interested in contributing or exploring the code from a developer perspective - and github makes it easy to have the `CONTRIBUTING.md` as the natural starting point (it is automatically mentioned when creating a PR for the first time, in the [community standards](https://github.com/IBM/aihwkit/community), etc). From that starting point, having the link to the development conventions would allow for easily discovering not only the test needs, but other pieces of information that are likely to be useful to that user group.

It's just a suggestion, though - having the information in `tests/README.md` is not extremely uncommon either (arguably a bit harder to incorporate into the overall documentation flow, and also blurring slightly the purpose of the `tests/` folder, which is mostly for the test themselves). The important part is that the information ends up in the repository one way or the other :+1:

examples/README.md
diego-plan9: Thanks! Centralizing the comments on the "where" in https://github.com/IBM/aihwkit/pull/374#discussion_r871142572

maljoras (dismissed):
Looks fine, except that Mapped layers are not mentioned. Are you planning to add those in a followup PR?

docs/source/using_pytorch.rst
```diff
@@ -21,21 +21,63 @@ Analog layers
...
+|                                                              | | composed of several input planes                  |                     |
++--------------------------------------------------------------+-----------------------------------------------------+---------------------+
+| :class:`~aihwkit.nn.modules.conv.AnalogConv3d`               | | Applies a 3D convolution over an input signal     | `nn.Conv3d`_        |
+|                                                              | | composed of several input planes                  |                     |
++--------------------------------------------------------------+-----------------------------------------------------+---------------------+
+| :class:`~aihwkit.nn.modules.linear_mapped.AnalogLinearMapped`| | Similar to AnalogLinearMapped but constrains the  | `nn.Linear`_        |
+|                                                              | | the maximal in and/or out dimension of an analog  |                     |
+|                                                              | | tile.                                             |                     |
```
maljoras, line 44: could we add here something like:
and will construct multiple tiles (as many as necessary to cover the weight matrix). Splitting, concatenation, and partial sum addition is done in digital.

docs/source/using_pytorch.rst
```diff
@@ -21,21 +21,63 @@ Analog layers
...
+~~~~~~~~~~~~~~~~
+
++-------------------------------------------------------------+-----------------------------------------------------+---------------------+
+| Analog Layer                                                | Description                                         | PyTorch Counterpart |
++=============================================================+=====================================================+=====================+
+| :class:`~aihwkit.nn.modules.rnn.rnn.AnalogRNN`              | | A modular RNN that uses analog tiles. Can take    |  | `nn.RNN`_        |
+|                                                             | | one of three types: AnalogLSTM, AnalogGRU, or     |  | `nn.LSTM`_       |
+|                                                             | | AnalogVanillaRNN                                  |  | `nn.GRU`_        |
++-------------------------------------------------------------+-----------------------------------------------------+---------------------+
+| :class:`~aihwkit.nn.modules.rnn.cells.AnalogVanillaRNNCell` | An Elman RNN cell with tanh or ReLU non-linearity.  | `nn.RNNCell`_       |
```
maljoras, line 75: "Elman"?

tests/README.md
```diff
@@ -0,0 +1,22 @@
+# IBM Analog Hardware Acceleration Kit: Unit Tests
+
+
+This project contains a unittest compatible test suite, that can be executed by any Python test runner. The recommended runner is pytest, which can be installed along with a number of other development tools used in aihwkit via:
+```
+pip install -r requirements-dev.txt 
+```
+
+All the tests and example do not have a main.py file. To run the full test suite simply do (from the command line):
```
maljoras, line 9: Maybe just remove the sentence "All the tests and example do not have a main.py file."

maljoras (changes requested):

maljoras (approved):

maljoras:
Thanks, @kaoutar55 !

PR #382, opened 2022-05-26
Change aihwkit version

kaoutar55:
## Related issues

None
## Description

Changed aihwkit to the correct version


diego-plan9 (approved):
Approving as bumping the version is needed - however, I'd recommend creating a new `0.6.1` release. Ideally, bumping the version should happen _before_ creating the release, and as a side effect the current `0.6.0` wheels contain a `0.5.1` version string (which is not critically problematic, but can cause some confusion).

In the past, that was the reason why usually the _last_ PR for a release is similar to #271:
* bumping the version number
* finalizing the changelog (replacing the `UNRELEASED` section with the new version and date). In this case, I'd also recommend adding a `Fixed` entry mentioning that the `0.6.0` release contains incorrect internal version information.

If making a release sounds good, it's fine to tackle it in this PR or in a separate one - happy to re-review if needed!

PR #383, opened 2022-05-26
Added an example factsheet for hw-aware training in documentation

kaoutar55:
## Related issues

None

## Description

Documentation enhancements


kaoutar55:
We will not include this factsheet in the public documentation.

PR #384, opened 2022-06-02
Multiple rnn

Jo7701:
This PR contains the inclusion of a new jupyter notebook (analog_training_rnns.ipynb) which walks through an example using an AnalogRNN model.


kaoutar55 (approved):

maljoras:
@kaoutar55 We should remove the data set here. It can be downloaded with torch

maljoras:
This seems to be a stale PR. Please reopen if necessary.

PR #388, opened 2022-06-24
Remapping weights functionality

maljoras:
## Related issues

When learning the out-scaling alpha, weight might become too small in relation to the weight max. In particular, when non-fixed clipping is used during HWA training.

Note: this function does only apply for hardware-aware training for analog inference. During analog in-memory training it should not be used. 

## Description

Added a `remap_weights` for `AnalogSequential` and `AnalogModuleBase` level. This function needs to be called by the user if used.

## Details
* also fixes an issue with setting contiguous weights


Fabio-83 (approved):
Many thanks @maljoras

PR #389, opened 2022-06-24
Fixing programming noise visualization and `program_analog_noise` method

maljoras:
* Fix: `program_analog_noise` would not change the weights alone (only together with `drift_analog_weights`)) in case no drift compensation was used
* Set drift compensation to optional
* `plot_weight_drift` would program the noise for each t_inference not clearly showing the programing versus read noise.    

## Details

* visualization now shows error bar only for read noise, programming noise can be seen from the onset.


Fabio-83 (approved):
Thanks @maljoras

PR #390, opened 2022-06-24
Concatenating updates in optimizer for performance reason

maljoras:
## Related issues

In the gradient trace for analog tiles, repeated gradients are appended to a list. Currently, the tile update function is called for each of these values. Thus, e.g. for LSTMs the optimizer loops through single updates for analog, so that analog training is slow (on GPUs). 

## Description
Here the update vectors are concatenated to the batch dimension, which will optimize the update computation in particular for CUDA


Fabio-83 (approved):
Thanks @maljoras

PR #391, opened 2022-06-24
LSTM cell with combined weights

Fabio-83:
## Description

Added a new LSTM cell which combines the weights of the LSTM network on a single Analog layer.

Added an example that explores the study done in https://www.frontiersin.org/articles/10.3389/frai.2021.699148/full on a 2-layer LSTM network trained with the War and Peace dataset.

## Details
In the example, there are 4 different RPU that can be used: a floating point tile to run the workload; an analog tile with ideal devices that use SGD for training; an analog tile with ideal and symmetric device that use SGD for training; an analog tile which uses ideal device with tiki-taka v2 algorithm that achieves better performance than SGD.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog], and this project adheres to
 ## Unreleased
 
 ### Added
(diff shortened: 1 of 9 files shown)


maljoras (approved):
Many thanks @Fabio-83 !

PR #392, opened 2022-06-24
fix to analog_summary when model is on cuda

Fabio-83:
## Description

The analog_info utils failed when the model was moved to cuda because of a discrepancy between model and input tensor device.

## Details

Now the model device is inferred from its parameter and the input tensor moved accordingly.


maljoras (approved):
Thank you, @Fabio-83 !

PR #393, opened 2022-06-24
Analog matrix-vector-product examples using analog tiles directly

maljoras:
## Description

For numerical computation, the auto grad functionality of pytorch is often not necessary. Here we add an example how to use the  `AnalogTile` class directly to implement a matrix-vector product in analog, without the pytorch modules.


Fabio-83 (approved):
Many thanks @maljoras

PR #399, opened 2022-07-27
Fixed: duplication of words

matifali:


maljoras (approved):
Many thanks, @matifali !

PR #402, opened 2022-08-07
Analog BERT

HCY-11:
## Description
Adds transformer modules from Huggingface in analog and allows for model loading/pretraining with Huggingface models.


maljoras:
Hi @HCY-11,

Many thanks for your contribution!  While your work is highly appreciated, most of the transformer script is based on the huggingface implementation and might be problematic to add to the toolkit from a legal perspective. Also it will introduce a lot of new dependencies to other packages. In particular, huggingface has been tricky in the past,  as it is changing fast without much notice or backward compatibility and introduces a lot of third party dependencies that are often not well compatible with e.g. power platforms. Adding these dependencies would make the aihwkit package a lot more difficult to maintain.  

That said, an analog transformer model is of course highly useful to have since it is a good match for the analog approach in general. Note that we have the `convert_to_analog` functionality (see [here](https://github.com/IBM/aihwkit/blob/master/src/aihwkit/nn/conversion.py#L44)) which translates all the linear layers and convolutions to equivalent analog layers for any torch model. So the most direct and simplest approach to support transformers from hugging face would be to simply to instantiate a hugging face model and then 

```python
analog_transformer = convert_to_analog(huggingface_transformer_model, rpu_config)
```
have you tried this with any hugging face model ? If so, did it work? If it works, would it be a valid approach to supporting transformers or would your implementation be needed nevertheless and why if the latter?

Many thanks!

HCY-11:
Ah, it seems I've misunderstood. `convert_to_analog` works fine and is pretty clearly the easiest way to go. I'll try creating some examples using that instead.

Issue #398, opened 2022-07-26
How to configure CUDA during installation?


matifali:
How can I configure CUDA during installation?
I have torch 1.12 installed with CUDA enabled.
But when I install aihwkit using
```
pip install -v aihwkit

```
It wants to install torch=1.8
I am using Nvidia RTX A5000, which does not support torch=1.8.

Also, is this the only requirement or do I need to change something else to enable CUDA support?

maljoras:
The package install does not support CUDA at the moment. For CUDA installation you need compile the code yourself. You can follow the development installation instructions . Which OS are you using? For windows you can try [this](https://aihwkit.readthedocs.io/en/latest/advanced_install.html#windows-using-conda-experimental)

In case of linux, you might want to follow [this](https://aihwkit.readthedocs.io/en/latest/developer_install.html) and use 
```make build_inplace_cuda``` which is a short form of 
 
```
python setup.py build_ext -j8 -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=TRUE --inplace  -DRPU_BLAS=MKL -DINTEL_MKL_DIR=${MKLROOT} -DUSE_CUDA=ON
```

**Comment by matifali** [comment]
I am using the `pytorch/pytorch:latest` docker image that comes with pytorch=1.12 and python=3.7
```
CMake Error in /home/atif/aihwkit/_skbuild/linux-x86_64-3.7/cmake-build/CMakeFiles/CMakeTmp/CMakeLists.txt:
  CUDA_ARCHITECTURES is empty for target "cmTC_d872e".

```
I am getting these errors

**Comment by maljoras** [comment]
You need to set the flag CUDA_ARCHITECTURES correctly. For instance by:
```
make build_cuda flags='-DRPU_CUDA_ARCHITECTURES="60"'
```
Not sure which architecture your are having.

**Comment by matifali** [comment]
Still same error
 ```CUDA_ARCHITECTURES is empty for target "cmTC_7572b"```

**Comment by maljoras** [comment]
That looks like that your CUDA library might not be installed properly? Or maybe try
```
make build_cuda flags='-DRPU_CUDA_ARCHITECTURES="86" -DCUDA_ARCHITECTURES="86"'
```
If that is not working make sure that it can find nvcc (you might have to set CUDA_HOME to the path where the cuda compilers are installed)
It would help if you posted the full error log.

**Comment by matifali** [comment]
This is the full log
```
[...]

Issue #297, opened 2021-10-30
matching cuda version to Pytorch version


ElinorG11:
Hello,

I'm using Linux Ubuntu 20.04, with python 3.8, Nvidia GeForce RTX 3070 and cuda version 11.4. I've installed aihwkit version 0.4.0, which requires PyTorch 1.8.1, and when running some of the git examples, I get the following error:

`aihwkit_env/lib/python3.8/site-packages/torch/cuda/__init__.py:104: UserWarning:
NVIDIA GeForce RTX 3070 with CUDA capability sm_86 is not compatible with the current PyTorch installation.
The current PyTorch install supports CUDA capabilities sm_37 sm_50 sm_60 sm_70.
If you want to use the NVIDIA GeForce RTX 3070 GPU with PyTorch, please check the instructions at https://pytorch.org/get-started/locally/

  warnings.warn(incompatible_device_warn.format(device_name, capability, " ".join(arch_list), device_name))
Traceback (most recent call last):
  File "./03_mnist_training.py", line 206, in <module>
    main()
  File "./03_mnist_training.py", line 195, in main
    model = create_analog_network(INPUT_SIZE, HIDDEN_SIZES, OUTPUT_SIZE)
  File "./03_mnist_training.py", line 98, in create_analog_network
    model.cuda()
  File "/home/elinor/aihwkit_env/lib/python3.8/site-packages/aihwkit/nn/modules/container.py", line 71, in cuda
    super().cuda(device)
  File "/home/elinor/aihwkit_env/lib/python3.8/site-packages/torch/nn/modules/module.py", line 491, in cuda
    return self._apply(lambda t: t.cuda(device))
  File "/home/elinor/aihwkit_env/lib/python3.8/site-packages/torch/nn/modules/module.py", line 387, in _apply
    module._apply(fn)
  File "/home/elinor/aihwkit_env/lib/python3.8/site-packages/torch/nn/modules/module.py", line 409, in _apply
    param_applied = fn(param)
  File "/home/elinor/aihwkit_env/lib/python3.8/site-packages/torch/nn/modules/module.py", line 491, in <lambda>
    return self._apply(lambda t: t.cuda(device))
  File "/home/elinor/aihwkit_env/lib/python3.8/site-packages/aihwkit/optim/context.py", line 80, in cuda
    self.analog_tile = self.analog_tile.cuda(device)
[...]

Issue #126, opened 2021-02-10
Issues on installation with latest CUDA toolkit and PyTorch


chaeunl:
## Description
Hello,
For several reasons, I am trying to update my CUDA toolkit and NVIDIA driver. As PyTorch only supports CUDA 11.0 version, I installed PyTorch 1.8.0 by wheel. But, during installation of aihwkit, I got an error message saying that:  
```
-- The CUDA compiler identification is unknown
-- Check for working CUDA compiler: /usr/bin/nvcc
CMake Error in /home/chaeunl/aihwkit/_skbuild/linux-x86_64-3.8/cmake-build/CMakeFiles/CMakeTmp/CMakeLists.txt:
  CUDA_ARCHITECTURES is empty for target "cmTC_6aef9".
```

A few months ago, I also tried with PyTorch 1.8.0 and it was successful. The differences are CUDA/cuDNN/NVIDIA driver version (previously, CUDA=11.1, cuDNN=8.0.4, NVIDIA driver=455.xx). Also, I tried with the same cmd: `python setup.py install -DUSE_CUDA=ON -DRPU_CUDA_ARCHITECTURES="80"`

I confirmed that cmake, pybind11, blas, and scikit-build satisfy the conditions described in installation page. Could you let me know how to fix it?

* Pytorch version: 1.8.0.dev20210208+cu110
* Package version: -
* OS: 18.04.2
* Python version: 3.8.5
* Conda version (or N/A):
* CUDA: 11.2
* cuDNN: 8.1
* NVIDIA driver: 460.39
* VGA: RTX 3090

diego-plan9:
Thanks for reporting as usual @chaeunl ! It seems PyTorch 1.8 is starting their release candidate publishing, but so far I could not find a non-conda GPU-enabled wheel for downloading, which would make troubleshooting easier: the good part is that it seems to be around the corner.

However, the
```
-- The CUDA compiler identification is unknown
```
line does look suspicious. Assuming you did a `make clean` to rule out artifacts of previous builds affecting the current one, would you know if under your previous setup, the cuda compiler was correctly identified by `cmake`?

maljoras:
Closing this issue. Please reopen if problem persists

Issue #56, opened 2020-10-30
CUDA 11 build compatibility


diego-plan9:
## Description and motivation

It seems that `CUB` is included along the CUDA Toolkit since version 11, which can cause issues during build (thanks @chaeunl for the valuable feedback and troubleshooting!):

```
$ python setup.py install -DUSE_CUDA=ON -DRPU_CUDA_ARCHITECTURES="80"
[ 14%] Built target cub
[ 50%] Built target RPU_CPU
[ 51%] Building CUDA object CMakeFiles/RPU_GPU.dir/src/rpucuda/cuda/bit_line_maker.cu.o
In file included from /usr/local/cuda-11.1/targets/x86_64-linux/include/thrust/system/cuda/detail/execution_policy.h:33:0,
                 from /usr/local/cuda-11.1/targets/x86_64-linux/include/thrust/iterator/detail/device_system_tag.h:23,
                 from /usr/local/cuda-11.1/targets/x86_64-linux/include/thrust/iterator/detail/iterator_facade_category.h:22,
                 from /usr/local/cuda-11.1/targets/x86_64-linux/include/thrust/iterator/iterator_facade.h:37,
                 from /.../aihwkit/_skbuild/linux-x86_64-3.8/cmake-build/cub-prefix/src/cub/cub/iterator/arg_index_input_iterator.cuh:48,
                 from /.../aihwkit/_skbuild/linux-x86_64-3.8/cmake-build/cub-prefix/src/cub/cub/device/device_reduce.cuh:41,
                 from /.../aihwkit/_skbuild/linux-x86_64-3.8/cmake-build/cub-prefix/src/cub/cub/cub.cuh:53,
                 from /.../aihwkit/src/rpucuda/cuda/bit_line_maker.cu:24:
/usr/local/cuda-11.1/targets/x86_64-linux/include/thrust/system/cuda/config.h:78:2: error: #error The version of CUB in your include path is not compatible with this release of Thrust. CUB is now included in the CUDA Toolkit, so you no longer need to use your own checkout of CUB. Define THRUST_IGNORE_CUB_VERSION_CHECK to ignore this.
#error The version of CUB in your include path is not compatible with this release of Thrust. CUB is now included in the CUDA Toolkit, so you no longer need to use your own checkout of CUB. Define THRUST_IGNORE_CUB_VERSION_CHECK to ignore this.
```

## Proposed solution

[...]

PR #403, opened 2022-08-08
chore:Added CUDA docker file and build script

matifali:
## Related issues

It is related to the following issues
#5 
#56
#126
#297 
#398

## Description

This pull request provides a working CUDA-enabled docker image.

1. A `CUDA.Dockerfile`
2. A `builddockerimageCUDA.sh` to build the image.

## Details
It extracts the CUDA_ARCHITECTURES from `nvidia-smi` automatically and build an image with CUDA enabled with the latest version of PyTorch and CUDA.
It is built on top of `nvidia/cuda`. 
**`CUDA_ARCH`**, **`CUDA_VER`**, and **`UBUNTU_VER`** cab be set as variables in the build script and Dockerfile.

For those who need the built image. They can download from docker-hub
```
docker pull matifali/aihwkit:86-cuda11.7ubuntu22.04
```
This is built for CUDA_ARCHITECTURE=86 (i.e, Any Nvidia card with compute capability of 8.6)


CUDA.Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+    echo "${USERNAME} ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/nopasswd
+
+# Change to your user
+USER ${USERNAME}
+WORKDIR /home/${USERNAME}
+
+# Install python packages as your user
+RUN pip install --upgrade pip
+RUN pip install pybind11 scikit-build
+RUN pip install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cu116
```
diego-plan9, line 54: Is there a way of making the pytorch repository used (ie. `.../cu116`) depend on the `CUDA_VER` argument? The [repository URLs](https://pytorch.org/get-started/locally/) seem to be relatively predictable, although not available for all available CUDA versions.

If it turns out to be not feasible or cumbersome, moving it to an argument and adding a note to the documentation stating that a specific version of pytorch is used should be enough - end goal is being able to accommodate different (and future) versions easily.

CUDA.Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+# use the k: https://developer.nvidia.com/cuda-gpus
+ARG CUDA_ARCH=86
+ARG NUM_CPU_THREADS=8
+RUN echo "Detected CUDA_ARCHITECTURE is = ${CUDA_ARCH}"
+# Build and install IBM aihwkit
+RUN make clean
+RUN python setup.py build_ext -j${NUM_CPU_THREADS} --inplace -DUSE_CUDA=ON -DRPU_CUDA_ARCHITECTURES="${CUDA_ARCH}" -DRPU_BLAS=OpenBLAS
+
+# Set paths of installation
+RUN echo 'export PYTHONPATH=$HOME/aihwkit/src' >> /home/${USERNAME}/.bashrc
```
diego-plan9, line 73: Can you replace these two lines with installing `aihwkit` via `pip`, rather than compiling the shared library and manually apppending it to the path? This would allow for the package to be available same as other python packages in a cleaner way.

Hopefully this (adapted) bit from the [docs](https://aihwkit.readthedocs.io/en/latest/advanced_install.html) would do the trick for passing cmake options to `pip install`:
```
pip install . --install-option="-DUSE_CUDA=ON" ...

Comment (diego-plan9) [reply CUDA.Dockerfile:73]: (On a more minor note, while editing the above, if you could get rid of the empty newlines at the end of the file would be great!)

File: CUDA.Dockerfile [hunk 3]
```diff
@@ -0,0 +1,75 @@
...
+RUN pip install --upgrade pip
+RUN pip install pybind11 scikit-build
+RUN pip install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cu116
+
+# Set path of python packages
+RUN echo 'export PATH=$HOME/.local/bin:$PATH' >> /home/${USERNAME}/.bashrc
+
+# clone the repo and change to the source directory
+RUN git clone https://github.com/IBM/aihwkit.git
+WORKDIR /home/${USERNAME}/aihwkit
```
diego-plan9, line 61: Instead of cloning the repository - can you copy the sources from the host, and use those instead? This would potentially allow for users to easily create new images based on something other than `master`, and also play nicer with eventual integrations.

builddockerimageCUDA.sh
```diff
@@ -0,0 +1,26 @@
...
+# Build Container
+docker build \
+--tag aihwkit:$CUDA_ARCH-cuda$CUDA_VER-ubuntu$UBUNTU_VER \
+--build-arg CUDA_ARCH=$CUDA_ARCH \
+--build-arg CUDA_VER=11.7 \
+--build-arg UBUNTU_VER=22.04 \
+--build-arg NUM_CPU_THREADS=$NUM_CPU_THREADS \
+--file CUDA.Dockerfile .
+
+echo "Docker image aihwkit:$CUDA_ARCH-cuda$CUDA_VER-ubuntu$UBUNTU_VER is built successfully"
```
diego-plan9, line 26: Can you add a reduced version of the script as documentation? Since the bulk of the script is the `docker build` command (and the temporary variables can also be evaluated inline), we can avoid the need for a separate file and at the same time provide users with some insight on how to tune the generation of the image if needed.

@maljoras might have better insights on its placement - my recommendation would be as a new block after [in-place-installation](https://github.com/IBM/aihwkit/blob/master/docs/source/advanced_install.rst#in-place-installation), ie.

```
...
Here, you need to make sure that the ``PYTHONPATH`` is set to the ``src``
sub-directory of the ahwkit base directory, e.g. by (when being in the base directory)::

    $ export PYTHONPATH=`pwd`/src:$PYTHONPATH


CUDA-enabled docker image
~~~~~~~~~~~~~~~~~~~~~~~~~

(some explanation on how to build the image, and a copy-pasteable `docker build` command)
...
```

diego-plan9 (commented):
Thanks @matifali ! Please see the inline comments in order to adjust slightly the focus of the image and cater to more usages: looks in a good direction already!

CUDA.Dockerfile
matifali: for now, there is no direct way to do that as PyTorch is not compiled for all Cuda versions.
Maybe we can give only two options
cuda 10 defaults to 10.2 (for cards that do not support CUDA 11)
and cuda 11 defaults to 11.6

CUDA.Dockerfile
matifali: This is resolved.

matifali (commented):
will update with required chnages

CUDA.Dockerfile
diego-plan9: Thanks!

CUDA.Dockerfile
diego-plan9: Thanks for looking into it: can you add a
```
ARG PYTORCH_PIP_URL=https://download.pytorch.org/whl/cu116
```
argument (along with a note on the documentation at https://github.com/IBM/aihwkit/pull/403#discussion_r943583257) on replacing this argument if using a specific cuda version? This would avoid having some extra logic and make it easier to future-proof.

diego-plan9 (commented):
Thanks @matifali! Can you also update the PR with the other inline comments?

CUDA.Dockerfile
matifali: Updated with mentioned changes

builddockerimageCUDA.sh
matifali: I think I have added all the necessary changes.

docs/source/advanced_install.rst
```diff
@@ -208,7 +208,29 @@ sub-directory of the ahwkit base directory, e.g. by (when being in the base dire
...
+--tag aihwkit:cuda \
+--build-arg USERNAME=${USER} \
+--build-arg USERID=(id -u $USER) \
+--build-arg GROUPID=(id -g $USER) \
+--build-arg CUDA_ARCH=${CUDA_ARCH} \
+--build-arg CUDA_VER=11.7 \
+--build-arg UBUNTU_VER=22.04 \
+--build-arg PYTORCH_PIP_URL=https://download.pytorch.org/whl/cu116 \
+--file CUDA.Dockerfile .
+```
```
diego-plan9, line 233:
```suggestion
As an alternative to a regular install, a CUDA-enabled docker image can also be
built using the ``CUDA.Dockerfile`` included in the repository.

In order to build the image, first identify the ``CUDA_ARCH`` for your GPU
using from `nvidia-smi` in your local machine::

    export CUDA_ARCH=$(nvidia-smi --query-gpu=compute_cap --format=csv | sed -n '2 p' | tr -d '.')
    echo $CUDA_ARCH

The image can be built via::

    docker build \
    --tag aihwkit:cuda \
    --build-arg USERNAME=${USER} \
    --build-arg USERID=(id -u $USER) \
    --build-arg GROUPID=(id -g $USER) \
    --build-arg CUDA_ARCH=${CUDA_ARCH} \
    --build-arg CUDA_VER=11.7 \
    --build-arg UBUNTU_VER=22.04 \
    --build-arg PYTORCH_PIP_URL=https://download.pytorch.org/whl/cu116 \
    --file CUDA.Dockerfile .

If building your image against a different CUDA or PyTorch version, please
ensure setting the build arguments accordingly.

```

Can you ensure there is a bit more context in the documentation, and using `.rst` syntax? Feel free to accept the suggested change directly!

diego-plan9 (commented):
Thanks for the prompt updates :tada: Just one item related to ensure the documentation renders properly and matches the rest of installation methods, and should be good to go!

docs/source/advanced_install.rst
matifali: looks good. Sorry I didn't format it well.

diego-plan9: No worries! `.rst` in particular is quite pesky :+1:

diego-plan9 (approved):
Thanks a lot @matifali !

maljoras:
Thanks a lot @matifali  for the contribution and @diego-plan9 for the review!

diego-plan9 (approved):

PR #412, opened 2022-08-15
Bugfix forward indexed for inference tile

### LINKED ISSUE #405: Weight modifier is not applied to convolution layers
## Description

The weight modifier setting of the `InferenceRPUConfig` is currently only affecting the `AnalogLinear` layers but ignored for convolutions. This has an effect on hardware-ware training with modifier noise or quantization.

### LINKED ISSUE #406: Drift compensation is not applied for conv layers
## Description

Drift compensation is no applied for convolutions, even though `drift_compensation` is set.

### LINKED ISSUE #407: Drift compensation is using a constant readout vector instead of one-hot vectors
## Description

Not clearly a bug, but currently the default drift compensation is using a constant input (`ones`) as read out vector (with a abs max calculation) which results to sub-optimal compensation in some DNNs.  

## Expected behavior

The drift compensation should use one-hot vectors to read out all weights and calculate the mean abs value as a reference value for compensation.

### LINKED ISSUE #408: Specifiy different RPU Config for some layers when converting to analog
## Description and motivation

When using `convert_to_analog` or convert_to_analog_mapped` one `RPUConfig` is applied to all layers in the DNN. However, sometimes some layers should behave differently, e.g. the first layer might be using `FloatingPointRPUConfig`. 

## Proposed solution

Use a user-defined function input in `convert_to_analog` that allows flexible setting of each layer by layer name.

maljoras:
## Related issues

closes #405, closes #406, closes #407, closes #408

## Description

Adds the `forward_indexed` override in the inference tile, which has affected the drift compensation and noise model for convolution layers. 

Also adds a number of small enhancements. 

## Details
* drift compensation now uses eye instead of ones
* analog state names with constants
* specific layer conversions (see #408)
* Fix print for mapped layers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ The format is based on [Keep a Changelog], and this project adheres to
 * Added ``analog_modules`` generator in ``AnalogSequential``. (\#410)
 * Added ``SKIP_CUDA_TESTS`` to manually switch off the CUDA tests
 * Enabling comparisons of ``RPUConfig`` instances. (\#410)
+* Specific use-defined function for layer-wise setting for RPUConfigs
+  in conversions. (\#412)
 
 ### Fixed
 * Analog_summary error when model is on cuda device. (\#392)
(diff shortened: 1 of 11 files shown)


Fabio-83 (approved):
Thanks @maljoras

PR #415, opened 2022-09-06
Unfolded convs

maljoras:
## Description

Internally an implicit unfolding was used for all convolutions. However, this is inefficient for CPUs and only intended for GPUs. Now, conv2D is unfolded for CPU by default, which speeds up the 2D convolutions on CPUs


Fabio-83 (approved):
Thanks @maljoras

PR #418, opened 2022-09-08
Expose stochastic rounding options for mixed precision 

maljoras:
## Description

Adds options to the `MixedPrecisionCompound` to use stochastic rounding in case of quantization of the gradients and/or activations (`stoc_round_x` and `stoc_round_d`)


Fabio-83 (approved):
Thanks a lot @maljoras

PR #420, opened 2022-09-25
fix (CUDA Docker): pre install `protobuf==3.20.2`

matifali:
Preinstall `prorobuf==3.20.2` to fix the installaton of aihwkit.  Without preinstalling `prorobuf`, `aihwkit` fails to install it as a dependency


maljoras (approved):

maljoras:
Many thanks, @matifali !

PR #423, opened 2022-09-29
Hardware-aware training for inference with mapped weights improvements and bugfixes

maljoras:
## Related issues
* LR rate was effected by the out scales for hardware-aware training 
* Remapping of weights during training 

## Description

Improved implementation and bugfixes of the weight scaling and mapping. Mainly changed / added the following:

1. Weight mapping scales are now separate from the learned out scales. The weight mapping scales are never learned but used for remapping (if enabled). The LR is correctly scaled so that the full update is on the analog weights
2. The forward / backward weight modifier is only called once per update not in the forward pass anymore. This makes sure that all multiple forward calls do not change the weight, which would result in wrong updates
3. Output scales can be learned either in scalar manner or channel wise (vector). 
4. Mapping parameters are slightly changed in names, to make them clearer and shorter
5. Remapping can be performed during inference, in which case the weights are insured to be mapped to gmax, ignoring potential output bound clipping issues. Remapping types are implemented and remapping is now done in C++ directly. 
6. Changed the standard input to the tile of the weight setters to tensors (from numpy)
7. Cleaner separation of `pre_update` and `post_update` operations in `BaseTile` (and same for forward / backward) for easy ovreloading of tile-level pre and post processing 

Please check the adapted [06 example](https://github.com/IBM/aihwkit/blob/master/examples/06_lenet5_hardware_aware.py) for how to use the new features

## Details
* See mapping.remap for new parameters
* `out_scaling_columnwise` and `weight_scaling_columnwise` new parameter names
* weight scaling omega is separate from the remapping (although it performs the same operation), since it is part of the base tile. In this case mapping is only done once at the beginning (when setting the weights). 
* improved the API documentation here and there
* Design change in the `set/get_weights`: Instead of having separate function for `_scaled` both functions on the tile level now have an `apply_weight_scaling` argument that will apply the weight scaling if any. Realistic versions are now separate methods. On the module level the weight scalings are applied by default whereas on the tile level it is off by default.
* `weight_scaling_omega` argument to layer modules is now depreciated


Fabio-83 (approved):
Many thanks @maljoras

PR #434, opened 2022-10-06
Legacy checkpoint read, fix checkpoint re-training issue 

maljoras:
## Related issues

* Legacy checkpoints loading did not load correctly with the new mapping scales in case of `InferenceRPUConfig`
* Re-training would not update the learning of the `out_scaling` factors 

## Description
Checkpoint load modified the parameter in the tile reconstruction, which however caused a mismatch in the `module._parameters` dictionary, which meant that `model.parameters()` picked the wrong references. 

This caused the model to not train after checkpoint loading, in case when `out_scaling_alpha` were set to be trained

Now the parameters are re-registered correctly for each load. 

## Details

* tests added to check `out_scaling_alpha` loading
* small fixes regarding the runners and experiments
* bump version to 0.6.2


Fabio-83 (approved):
Thank you @maljoras

PR #456, opened 2022-12-09
Backend version update

### LINKED ISSUE #290: Tool for fitting device data to models in AIHWKIT
## Description and motivation

Device models in AIHWKIT have a number of parameters and it is often not quite obvious how to best fit a given device data (e.g. pulse response curves) to a given model. 

## Proposed solution

Make a script that automatically fits the (some of the) parameters to a given device data (e.g. similar to the traces generated by `plot_device`)

### LINKED ISSUE #291: Tool/tutorial for computing the symmetry points of a given device and setting up a TransferCompound
## Description and motivation

`TransferCompound` used for the Tiki-taka learning rule requires the symmetry point of each device to be set near zero. We have a `ReferenceUnitCell`l to support arbitrary subtraction, however, it is not immediately obvious how to do that as it requires setting of hidden parameter settings and the symmetry points are also not computed numerically automatically.   

## Proposed solution

Make a tool/script that computes and sets symmetry point to the reference devices for arbitrary device model settings.

### LINKED ISSUE #346: Bump mypy version

maljoras:
## Related issues

closes #346, closes #291, closes #290

## Description

Many updates and fixes in particular to the RPUCuda backend, including 

* Supporting analog tiles moving from GPU  to CPU
* non-blocking option for tile computation in CUDA 
* Input range learning for inference 
* bump pylint / mypy / pycodestyle versions  
* Added support for N-degree polynomial weight noise models for hardware-aware training for analog inference .
* Reset method for tiles on updated device
* Many more non-linearties for the analog mat-vec including multiple passes (positive / negative)
* Hidden weight write noise fixes
* Added a `remap_weights` for `AnalogSequential` and `AnalogModuleBase` level.
* Improved memory management of CUDA tiles
* `PowStepReference` and `SoftBoundsReference` device, with option to subtract symmetry point (see #291) 
* Fitting utility  (#290)
 
### Input range learning 

Learning of the input clipping bound.  
* Gradient depends on clipping of input as well as output 
* Decay
* adjustable parameters in `InputRangeParameters`

Yields very good results (see example 06)
image

### non blocking

* `CudaTiles` can be made no blocking by setting `tile.non_blocking = True`


kaoutar55 (approved):

kaoutar55:
Thanks @maljoras! can we merge this before the 0.7.0 release?

PR #469, opened 2023-01-31
Update CUDA.Dockerfile to support v0.7.0

matifali:
## Related issues

With the 0.7.0 version released, the `protobuf` minimum version requirement is changed, and there is a need to update the CUDA Dockerfile.

## Description

bump the `protobuf` version in `CUDA.Dockerfile`

### DIFF
--- a/CUDA.Dockerfile
+++ b/CUDA.Dockerfile
@@ -1,9 +1,9 @@
 # Build argumnets
-ARG CUDA_VER=11.7
+ARG CUDA_VER=11.8.0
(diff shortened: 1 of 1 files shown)


maljoras:
@matifali great, many thanks for the update!

maljoras (approved):

matifali:
What do you think about publishing an official docker image? I can help create an automated workflow to build and push if you agree and provide the credentials for pushing to a registry.
@maljoras

PR #471, opened 2023-02-01
BERT example with distributed compute

HCY-11:
## Related issues

Hardware-aware training with the current example
takes very long to run since it only uses one GPU.

## Description

This is a start for parallelizing the example by using torch.multiprocessing. Letting the Trainer know what the local rank is causes training to be in distributed compute. Currently, it appears that training can be completed, but once the example moves onto inference, it throws a SIGBUS error.


maljoras:
@HCY-11 many thanks! Can you include the fixes and corrections I made for the original examples 24? See PR #488

maljoras:
Closing this due to inactivity. Please reopen if needed

PR #495, opened 2023-03-27
Update .travis.yml

kkvtran:
Add compile and test stanza for python 3.11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,6 +83,11 @@ jobs:
     <<: *job_compile_common
     stage: Test multiple python versions
     python: "3.10"
+    
+   - name: "Compile and test. Python 3.11"
+    <<: *job_compile_common
+    stage: Test multiple python versions
+    python: "3.11" 
 
   - name: "Build wheel for Python 3.9, 3.10, 3.11 on Linux x86_64"
     os: linux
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -83,6 +83,11 @@ jobs:
     <<: *job_compile_common
     stage: Test multiple python versions
     python: "3.10"
+    
+   - name: "Compile and test. Python 3.11"
```
diego-plan9, line 87: Can you remove one space at the beginning of the line? YAML is a bit picky about indenting - and the missing space is causing the file not to be a valid `.yml` file, which in turn prevents travis from running:

```
  - name: "Compile and test. Python 3.11"
    <<: *job_compile_common
    stage: Test multiple python versions
    python: "3.11"
```

diego-plan9 (changes requested):

.travis.yml
kkvtran: Done.

diego-plan9: The travis log is at https://github.com/IBM/aihwkit/runs/12333481252 - for 030fdef, as the .yaml was malformed, no travis run took place (hence no link to the build).

PR #496, opened 2023-03-28
Update .travis.yml

kkvtran:
## Description

Build latest aihwkit wheels with python 3.8 and 3.9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,22 +74,17 @@ jobs:
       - make pylint
       - make mypy
 
+  - name: "Compile and test. Python 3.8"
+    <<: *job_compile_common
+    stage: Test multiple python versions
+    python: "3.8"
+
   - name: "Compile and test. Python 3.9"
     <<: *job_compile_common
     stage: Test multiple python versions
     python: "3.9"
-
-  - name: "Compile and test. Python 3.10"
-    <<: *job_compile_common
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -74,22 +74,17 @@ jobs:
...
     <<: *job_compile_common
     stage: Test multiple python versions
-    python: "3.9"
+    python: "3.8"
 
-  - name: "Compile and test. Python 3.10"
+  - name: "Compile and test. Python 3.9"
     <<: *job_compile_common
     stage: Test multiple python versions
-    python: "3.10"
```
maljoras, line 85: Why do you want to test it also for 3.9 only?  Can we stay on 3.10 for that?

.travis.yml
```diff
@@ -74,22 +74,17 @@ jobs:
       - make pylint
       - make mypy
 
-  - name: "Compile and test. Python 3.9"
+  - name: "Compile and test. Python 3.8"
     <<: *job_compile_common
```
maljoras, line 78: Why do we need to change this here? Only the wheel can change, no?

maljoras (changes requested):

kaoutar55 (approved):

.travis.yml
kkvtran: Done.

PR #546, opened 2023-10-18
enable fusion g_max configurable

kkvtran:
## Description

Some minor changes for pylint and travis build.

### DIFF
--- a/src/aihwkit/inference/calibration/calibration.py
+++ b/src/aihwkit/inference/calibration/calibration.py
@@ -119,9 +119,12 @@ def calibrate_input_ranges(
 
     Only tiles that support and have enabled input range learning will
     be calibrated. If noise management is turned on an error is
-    raised.  Note: this implementation transiently registers a new `forward_pre_hook`
-    on the analog tile level. It assumes that the user has not defined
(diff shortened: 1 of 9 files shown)


src/aihwkit/inference/calibration/calibration.py
```diff
@@ -119,7 +119,9 @@ def calibrate_input_ranges(
 
     Only tiles that support and have enabled input range learning will
     be calibrated. If noise management is turned on an error is
-    raised.
+    raised.  Note: this implementation transiently registers a new `forward_pre_hook`
```
maljoras, line 122: Can you make a newline before "Note:" and after like:
```
[..] raised.

Note: 
      This implementation transiently registers a new `forward_pre_hook` ...
      [and so on with spaces at the beginning]
```
In this way the docstring will be interpreted as Note with SPHINX

maljoras (changes requested):

src/aihwkit/inference/calibration/calibration.py
kkvtran: OK.  will do.

maljoras (approved):

PR #551, opened 2023-10-25
chore: update CUDA.Dockerfile to use multistage builds

matifali:
1. Updates the Dockerfile to use a multistage build
2. Use Intel MKL instead of OpenBLAS
3. Reduced the Final image size from 17+ GB to ~12 GB

### DIFF
--- a/CUDA.Dockerfile
+++ b/CUDA.Dockerfile
@@ -1,8 +1,9 @@
 # Stage 0: Intel MKL
+ARG CUDA_VERSION=11.8.0
 FROM intel/oneapi-basekit AS mkl-env
 
 # Stage 1: Build dependencies
-FROM nvidia/cuda:11.8.0-devel-ubuntu22.04 AS build-env
+FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu22.04 AS build-env
 
 # Install as root
 USER root
@@ -36,6 +37,10 @@ RUN groupadd -g ${GROUPID} ${USERNAME} && \
 # Copy MKL libraries from mkl-env
 COPY --from=mkl-env /opt/intel/oneapi/mkl /opt/intel/oneapi/mkl
 
+# Copy IBM aihwkit and build
+COPY . /aihwkit
+RUN chown -R ${USERNAME}:${USERNAME} /aihwkit
+
 # Change to your user
 USER ${USERNAME}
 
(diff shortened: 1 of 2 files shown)


CUDA.Dockerfile
```diff
@@ -1,10 +1,8 @@
-# Build argumnets
-ARG CUDA_VER=11.8.0
-ARG UBUNTU_VER=22.04
```
matifali, line 3: I have removed these ARGs. We can only build with 11.8.0, and aihwkit does not build with CUDA 12.x versions.

CUDA.Dockerfile
```diff
@@ -40,26 +93,20 @@ RUN groupadd -g ${GROUPID} ${USERNAME} && \
...
+# Install Python packages
+RUN pip install --no-cache-dir --no-warn-script-location \
+    matplotlib
 
-# Copy the source code inside to image and change to the source directory
-COPY . ./aihwkit
-WORKDIR /home/${USERNAME}/aihwkit
+# Environment variables
+ENV PATH /home/${USERNAME}/.local/bin:$PATH
+ENV LD_LIBRARY_PATH /opt/intel/oneapi/mkl/latest/lib/intel64:${LD_LIBRARY_PATH}
```
matifali, line 109: This makes sure that MKL libs are in the path.

docs/source/advanced_install.rst
```diff
@@ -257,9 +257,6 @@ The image can be built via::
     --build-arg USERID=(id -u $USER) \
     --build-arg GROUPID=(id -g $USER) \
     --build-arg CUDA_ARCH=${CUDA_ARCH} \
-    --build-arg CUDA_VER=11.8.0 \
-    --build-arg UBUNTU_VER=22.04 \
-    --build-arg PYTORCH_PIP_URL=https://download.pytorch.org/whl/cu117 \
```
matifali, line 262: Removed as the image is hard-coded to cuda 11.8.

CUDA.Dockerfile
matifali: Edit: It builds fine, but while running requires some `cudablas11.so` files that are nit available with CUDA 12.1.

matifali: I will update

matifali: pushed a fix

matifali:
Hi @maljoras it is ready for review.

maljoras (approved):
Many thanks @matifali !

PR #556, opened 2023-11-22
chore: update CUDA.Dockerfile to use CUDA 12.2.2 by default.

matifali:
## Description

1. Update CUDA.Dockerfile to use CUDA 12.2.2 by default
2. Remove the PyTorch install URL to always default to the latest CUDA version.
3. Add a link to the instructions in README.md


maljoras (approved):
Many thanks, @matifali !

PR #558, opened 2023-11-29
update to use ubuntu 20.04 focal

kkvtran:
## Details

Update to use Ubuntu focal version 20.04.


maljoras (approved):
Many thanks @kkvtran!

PR #563, opened 2023-12-11
Number formats and bugfixes

maljoras:
## Related issues

Make ready for RPU tiles with different number format (fp16 / double).   

## Description

The C++ code and python code is adapted to use different simulation number format. For that a new `runtime.data_type` field is added to the `RPUConfig`. Once constructed the tile  can be used to run the specified input data types. 

However: full FP16 support is currently experimental (turned off by default) and only available for CUDA >=12.  

There is also an option to store the parameter of pulsed devices into FP16 which reduces the memory for analog training.  

Additional, an C++ aihwkit extension module for additional C++ code is added (optionally). Currently, the extension module only has a fake FP32 format conversion operator (for variable mantissa / exponent with saturation and rounding).   

All options are experimental and turned off by default and need to be turned off when compiling. 

Additional bugfixes and changes:
- convert to digital of an analog model added
- bugfixes / feature additions to the specialized in-memory learning rules 
- AnalogWrapper improvements (now a direct subclass)
- some reshuffling of the configs/ parameter files to reduce python file size 
- RNN analog linear layer more flexible
- custom tiles improvements
- `RPUConfig` base classes changes


examples/06_lenet5_hardware_aware.py
```diff
@@ -365,62 +367,65 @@ def inference_phase(t_inference_times, model, criterion, validation_data):
...
-# Train the model
-results = training_phase(analog_model, crit, opt, training_data, valid_data)
-
-
-# Test model inference over time
-t_inference_lst = [0.0, 1.0, 20.0, 1000.0, 1e5, 1e7]
-inference_error, _ = inference_phase(t_inference_lst, analog_model, crit, valid_data)
-plot_results(*results, t_inference_lst, inference_error)
+if __name__ == "__main__":
+    # Make sure the directory where to save the results exist.
```
kaoutar55, line 371: exists

kaoutar55 (approved):
Thanks @maljoras

PR #570, opened 2023-12-20
Add readthedocs configuration file

diego-plan9:
## Related issues

N/A

## Description

Add a configuration file for readthedocs, in order to re-enable the building of documentation.

## Details

Since a few months ago, it seems `readthedocs` requires a config file to be present in the repo in order to render the documentation, which in turned caused the documentation not to be automatically updated since 776242d4. Thanks to @kaoutar55 for raising the issue.

Extra notes:
* In the process, I added a missing dependency to the doc requirements.
* I tested locally by invoking `make doc` under a 3.11 environment - there is a small number of  warnings (documents not included in toctrees, etc), but felt outside the scope of the PR (hopefully this is only for re-enabling doc rendering)
* the real test will be once the PR is merged, and the readthedocs integration is (hopefully) triggered

References:
* announcement: https://blog.readthedocs.com/migrate-configuration-v2/
* config file documentation: https://docs.readthedocs.io/en/stable/config-file/v2.html

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,35 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.12"
+    # You can also specify other tool versions:
+    # nodejs: "20"
+    # rust: "1.70"
+    # golang: "1.20"
+
+# Build documentation in the "docs/" directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+  # You can configure Sphinx to use a different builder, for instance use the dirhtml builder for simpler URLs
+  # builder: "dirhtml"
+  # Fail on all warnings to avoid broken references
+  # fail_on_warning: true
+
+# Optionally build your docs in additional formats such as PDF and ePub
+# formats:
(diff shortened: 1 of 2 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,23 @@
...
+sphinx:
+  configuration: docs/source/conf.py
+
+# Optional but recommended, declare the Python requirements required
+# to build your documentation
+# See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
+python:
+  install:
+    - requirements: docs/requirements.txt
+    - requirements: requirements.txt
```
diego-plan9, line 23: Added this include so the package dependencies are installed, and mostly for the API docs - without it, most of the pages would not render as they would not be able to `import torch`, etc. The dependencies were removed from `doc/requirements.txt` at #456 - hopefully this will accomplish the same goals.

.readthedocs.yaml
```diff
@@ -0,0 +1,23 @@
...
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
```
diego-plan9, line 11: I went a bit conservative with `3.11` instead of `3.12`, as I did not have a good environment to test it. As long as the pinned version of `sphinx` in `docs/requirements.txt` works, it should be safe to bump.

PR #585, opened 2023-12-29
Render aihwkit.simulator.rpu_base in readthedocs

### LINKED ISSUE #9: Document aihwkit.simulator.rpu_base in readthedocs
## Description and motivation

Currently, `aihwkit.simulator.rpu_base` is not properly displayed in [the API reference](https://aihwkit.readthedocs.io/en/latest/api_reference.html). This is due to not being able to compile the simulator in the readthedocs build environment, as it does not allow the same flexibility when installing external packages (and at the moment the bindings need to be compiled in order to produce its documentation correctly).

It would be great to have online documentation for that module as well, in the same way as the rest of the modules.

We might have some luck using `.pyi` stubs - even if they seem not fully supported yet in sphinx (https://github.com/sphinx-doc/sphinx/pull/4824), there are pointers for other options such as https://github.com/readthedocs/sphinx-autoapi.

diego-plan9:
## Related issues

Closes #9 

## Description

Take advantage of the new `readthedocs` config file features in order to (hopefully) have `aihwkit.simulator.rpu_base` displayed in the readthedocs documentation, by:
* installing required non-Python requirements via `apt`
* installing the package using `pip`
* cleaning up some extra tweaks in the docs config that were needed due to the exclusion of the module under readthedocs

## Details

Please note I have **not** been able to test the changes, as currently they will only be exercised if the PR is merged (ie. after a commit to `master`, and reviewing the readthedocs build log) ... and it might take a few rounds of iterating over the changes in order to get it to work (with the risk of the documentation being broken until fixed/reverted).

It should be possible to enable [building the docs on PRs](https://docs.readthedocs.io/en/stable/pull-requests.html) to ease iterating on the changes, although this might also require the PR to be issued from a branch in the `aihwkit` repo rather than from a fork. I'd suggest trying that route instead of merging-and-reverting-if-fails - please feel free to hijack the PR as needed to your liking!

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,8 +8,11 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.11"
-
+    python: "3.10"
+  apt_packages:
+    - libopenblas-dev
+    - libpython-dev
+    - gcc
 # Build documentation in the "docs/" directory with Sphinx
 sphinx:
   configuration: docs/source/conf.py
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -24,3 +24,8 @@ python:
   install:
     - requirements: docs/requirements.txt
     - requirements: requirements.txt
+    - method: pip
+      path: .
+      extra_requirements:
+        - visualization
+        - fitting
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -48,6 +48,7 @@ API Reference
    aihwkit.simulator
    aihwkit.simulator.configs
    aihwkit.simulator.presets
+   aihwkit.simulator.rpu_base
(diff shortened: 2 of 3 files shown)


.readthedocs.yaml
```diff
@@ -8,8 +8,11 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.11"
-
+    python: "3.10"
```
diego-plan9, line 11: The downgrade was as `ubuntu-22.04` provides [Python 3.10](https://packages.ubuntu.com/jammy/python3-dev) , and due to trying to match `libpython-dev`. I'm not sure if readthedocs will use ubuntu's 3.10 version, but at least trying to minimize the risk of version mismatch.

maljoras (approved):
Many thanks, @diego-plan9!

PR #581, opened 2023-12-28
Docker mypy

maljoras:
## Related issues

closes #566 

## Description

Stubgen was not found for the docker build since mypy was not installed. 

## Details

In future it would be best to install the requirements.txt directly from the docker build file

PR #586, opened 2024-01-02
Fix readthedocs builds

diego-plan9:
## Related issues

#585 

## Description

Unfortunately, as hinted during #585, the initial build with `rpu_base` enabled has failed, due to an issue with the name of one `apt` dependency. This PR resolves the failure - and additionally, adds more tweaks in order to get a passing build.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -5,3 +5,5 @@ myst-parser==2.0.0
 # Dependencies for extras in `setup.py` needed for documentation.
 matplotlib>=3.0
 lmfit
+# Dependencies needed for building the package
+mypy==0.991
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -13,7 +13,7 @@
 import os
 import sys
 
-sys.path.insert(0, os.path.abspath("../../src"))
+# sys.path.insert(0, os.path.abspath("../../src"))
 
 
 # -- Project information -----------------------------------------------------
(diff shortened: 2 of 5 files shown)


docs/requirements.txt
```diff
@@ -5,3 +5,5 @@ myst-parser==2.0.0
 # Dependencies for extras in `setup.py` needed for documentation.
 matplotlib>=3.0
 lmfit
+# Dependencies needed for building the package
+mypy==0.991
```
diego-plan9, line 9: Small throwback to #581 - it seems that `mypy` is in practice a build dependency currently (due to the need for `stubgen`). I was not sure whether it should be promoted to a proper dependency, so opted for ensuring it becomes a "doc" dependency at least (as this file is installed by readthedocs).

docs/source/conf.py
```diff
@@ -13,7 +13,15 @@
...
-sys.path.insert(0, os.path.abspath("../../src"))
+# -- Options specific to readthedocs -----------------------------------------
+
+on_readthedocs = os.environ.get("READTHEDOCS") == "True"
+if not on_readthedocs:
+    # If not invoked from the `readthedocs` build environment, use the source
+    # files instead of assuming the package is installed, and mock `rpu_base`.
+    import sys
+    sys.path.insert(0, os.path.abspath("../../src"))
+    autodoc_mock_imports = ["aihwkit.simulator.rpu_base"]
```
diego-plan9, line 24: Small re-hash of some bits that were removed at #585 . In order to correctly produce the `.html` files for `rpu_base`, `autosummary` needs to work on the installed package (as, if working on just the `.py` files without installing, it will not be able to resolve the reference). If not inside the readthedocs environment, I aimed at preserving the previous behaviour.

maljoras (approved):

Issue #614, opened 2024-02-02
`WeightModifierType.DISCRETIZE` does not work on CUDA


jubueche:
## Description
I wrote a test checking the discretisation behaviour of the discretisation weight modifier and it fails if everything is placed on CUDA.

Note: The C++ version also does not work. One can add
`T res_ = res > (T)1.0 ? 2 / res : 2 * res;`
after [this line](https://github.com/IBM/aihwkit/blob/4b0ad90ea8f5de72d4c6454450ac4787e100a3db/src/rpucuda/weight_modifier.cpp#L87).

## How to reproduce

MWE:
```
from typing import Union
import torch
from torch import allclose, randn
from aihwkit.nn import AnalogLinear
from aihwkit.simulator.configs import (
    TorchInferenceRPUConfig,
    InferenceRPUConfig,
    WeightModifierType,
    NoiseManagementType,
    BoundManagementType,
)


def test_weight_modifier(
    modifier_res: float, wm_type: WeightModifierType = WeightModifierType.DISCRETIZE
):
    """
    Test correctness of discretization.
    """
    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
    # device = torch.device("cpu")

    def populate_rpu(
        rpu_config: Union[TorchInferenceRPUConfig, InferenceRPUConfig],
        modifier_res: float,
        wm_type: WeightModifierType,
    ):
        rpu_config.forward.bound_management = BoundManagementType.NONE
        rpu_config.forward.noise_management = NoiseManagementType.NONE
        rpu_config.noise_model = None
        rpu_config.modifier.type = wm_type
        rpu_config.modifier.res = modifier_res
        rpu_config.modifier.sto_round = False
        rpu_config.forward.inp_res = -1
        rpu_config.forward.out_res = -1
        rpu_config.forward.inp_bound = -1
        rpu_config.forward.out_bound = -1
        rpu_config.forward.out_noise = 0.0
        rpu_config.forward.is_perfect = True
        return rpu_config

    torch.manual_seed(0)
    tile_weights = torch.randn(256, 256).to(device)
    inp = randn((1, 256)).to(device)

    rpu_config_torch = populate_rpu(TorchInferenceRPUConfig(), modifier_res, wm_type)
    rpu_config = populate_rpu(InferenceRPUConfig(), modifier_res, wm_type)

    linear = AnalogLinear(in_features=256, out_features=256, bias=False, rpu_config=rpu_config)
[...]

PR #618, opened 2024-02-19
Added WeightModifierType.DISCRETIZE_PER_CHANNEL

jubueche:
## Related issues

No related issue for this new feature, but in this PR, I also want to address [this issue](https://github.com/IBM/aihwkit/issues/614).

## Description

I added a new `WeightModifierType.DISCRETIZE_PER_CHANNEL` type.

## Details

So far, we only have `WeightModifierType.DISCRETIZE` which discretizes the weight per-tensor. However, that induces too much quantization noise, so people resort to per-channel discretization, which is what I have implemented.
I also implemented a test that checks the correctness against a by-hand quantization in pure torch.

### DIFF
--- a/src/rpucuda/cuda/weight_modifier_cuda.cu
+++ b/src/rpucuda/cuda/weight_modifier_cuda.cu
@@ -14,19 +14,6 @@
 #include "cuda_math_util.h"
 #include "weight_modifier_cuda.h"
 
-#define cudaSafeCall(call)                                                     \
-  do {                                                                         \
-    cudaError_t err = call;                                                    \
-    if (cudaSuccess != err) {                                                  \
-      std::cerr << "CUDA error in " << __FILE__ << "(" << __LINE__             \
-                << "): " << cudaGetErrorString(err);                           \
-      exit(EXIT_FAILURE);                                                      \
-    }                                                                          \
-  } while (0)
-template <class T> __host__ __device__ T min_(T a, T b) {
-  return !(b < a) ? a : b;
-};
-
 namespace RPU {
 
 #define RPU_WM_KERNEL_LOOP(STOCH_IF, BODY)                                                         \
--- a/src/rpucuda/cuda/weight_modifier_cuda.cu
+++ b/src/rpucuda/cuda/weight_modifier_cuda.cu
@@ -14,6 +14,19 @@
 #include "cuda_math_util.h"
 #include "weight_modifier_cuda.h"
 
+#define cudaSafeCall(call)                                                     \
+  do {                                                                         \
+    cudaError_t err = call;                                                    \
+    if (cudaSuccess != err) {                                                  \
+      std::cerr << "CUDA error in " << __FILE__ << "(" << __LINE__             \
+                << "): " << cudaGetErrorString(err);                           \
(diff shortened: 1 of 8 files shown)


jubueche:
@maljoras the implementation already works for the CPU part, but I still need to implement it in CUDA for the GPU.
As written in the description, we can also use this to update the design of specifying res for this modifier.
@kkvtran maybe you can help with updating the documentation once the technical part is done and reviewed by @maljoras .

src/rpucuda/cuda/weight_modifier_cuda.cu
```diff
@@ -425,12 +495,19 @@ void WeightModifierCuda<T>::apply(
...
   T *amax = nullptr;
   if (wmpar.rel_to_actual_wmax && wmpar.type != WeightModifierType::Copy) {
-    if (!amaximizer_) {
-      amaximizer_ = RPU::make_unique<Maximizer<T>>(
-          context_, wmpar.copy_last_column ? (size_ - d_size_) : size_, true);
+    if (wmpar.type == WeightModifierType::DiscretizePerChannel) {
+      cudaMalloc((void**)&amax, d_size_ * sizeof(T));
+      int threadsPerBlock = 256; // what is better for this?
+      int blocksPerGrid = (x_size_ * d_size_ + threadsPerBlock - 1) / threadsPerBlock;
+      maxAbsAlongDimension<<<blocksPerGrid, threadsPerBlock, threadsPerBlock * sizeof(T)>>>(weights, x_size_, d_size_, amax, 0);
```
jubueche, line 502: I compute the max(abs(.)) per column here

src/rpucuda/cuda/weight_modifier_cuda.cu
```diff
@@ -462,6 +539,17 @@ void WeightModifierCuda<T>::apply(
     }
     break;
   }
+  case WeightModifierType::DiscretizePerChannel: {
+    if (wmpar.res > (T)0.0) {
+      kernelModifyWeightsDiscretizePerChannel<T><<<nblocks, nthreads, 0, s>>>(
+          size_, d_size_, wmpar.copy_last_column, new_weights, weights, wmpar.res, wmpar.sto_round,
+          wmpar.assumed_wmax, amax,
+          wmpar.sto_round ? context_->getRandomStates(nblocks * nthreads) : nullptr);
+      cudaFree(amax);
```
jubueche, line 548: I do the discretization and free the memory again that was used for the per-column amax.

tests/test_cuda_discretize_per_channel.py
jubueche, line 1: This script passes, but the test does not pass. The test has some weights that are zero.

jubueche: @maljoras after the weight modification of discretize per column, some weights are zero and therefore my test is not passing. This has something to do with CUDA, but I don't know how to fix it. Do you have an idea?

maljoras:
Is this really needed? That was the reason of having the weight remapping. So essentially, when you use channel wise weight remapping, you will get channel wise discretion with the original weight modifier for free

jubueche:
You're right, I forgot about the channel-wise re-mapping.

Issue #623, opened 2024-03-08
Training with half-precision doesn't work for the torch tile or CUDA bindings


coreylammie:
## Description

Training with half-precision doesn't work for the torch tile or CUDA bindings, e..g, when `with torch.autocast(device_type="cuda", dtype=torch.bfloat16):` is used in conjunction with `rpu_config.runtime.data_type = RPUDataType.HALF`.

## How to reproduce

Convert a model with either `TorchInferenceRPUConfig()` or `TorchInferenceRPUConfig()`, specify `rpu_config.runtime.data_type = RPUDataType.HALF`, and use `with torch.autocast(device_type="cuda", dtype=torch.bfloat16):` in the training loop.

## Expected behavior

https://aihwkit.readthedocs.io/en/latest/api/aihwkit.simulator.parameters.enums.html#aihwkit.simulator.parameters.enums.RPUDataType infers that this is supported.

jubueche:
I see only `fp16` in the Docs that you linked, but you are doing `bf16`. Does that explain it? And what exactly fails? Can you give a MWE?

coreylammie:
@jubueche no, neither work. The error is different for `bfloat16`, `float16`, and for the `torch tile` and `CUDA bindings`.

MWE:
```
import torch
import torch.nn as nn
import torch.nn.functional as F
from torchvision import datasets, transforms
from aihwkit.simulator.configs import InferenceRPUConfig, TorchInferenceRPUConfig
from aihwkit.nn.conversion import convert_to_analog
from aihwkit.optim import AnalogSGD
from aihwkit.simulator.parameters.enums import RPUDataType

device = torch.device("cuda" if torch.cuda.is_available() else "cpu")

class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.conv1 = nn.Conv2d(1, 32, 3, 1)
        self.conv2 = nn.Conv2d(32, 64, 3, 1)
        self.dropout1 = nn.Dropout(0.25)
        self.dropout2 = nn.Dropout(0.5)
        self.fc1 = nn.Linear(9216, 128)
        self.fc2 = nn.Linear(128, 10)

    def forward(self, x):
        x = self.conv1(x)
        x = F.relu(x)
        x = self.conv2(x)
        x = F.relu(x)
        x = F.max_pool2d(x, 2)
        x = self.dropout1(x)
        x = torch.flatten(x, 1)
        x = self.fc1(x)
        x = F.relu(x)
[...]

PR #625, opened 2024-03-08
Fix the support of different dtypes for the torch model

jubueche:
## Related issues

Issue #623 

## Description

Changed the .to function mostly.

## Details
- Verification on GPU needed
- A tests needs to be added that verifies the output correctness against pure torch.


jubueche:
Added tests for the torch version and verified that it is working on a A100 GPU.

jubueche:
Also, I added a line in the helper.py to make `make mypy` pass with a newer mypy` version.

src/aihwkit/simulator/tiles/module.py
```diff
@@ -227,23 +224,31 @@ def to(self, *args: Any, **kwargs: Any) -> "TileModule":
...
-            device = torch_device(args[0])
+        device = kwargs.get("device")
+        dtype = kwargs.get("dtype")
+        for arg in args:
+            if isinstance(arg, bool):
+                continue
+            if isinstance(arg, torch_dtype):
+                dtype = arg
+            if isinstance(arg, (str, torch_device)):
+                device = torch_device(arg) if isinstance(arg, str) else arg
```
maljoras, line 236: What was the problem here? Not sure whether the suggested solution is better than the original version. Does kwargs.get("device") does not throw an error when it is not found for instance?

src/aihwkit/simulator/tiles/module.py
```diff
@@ -227,23 +224,31 @@ def to(self, *args: Any, **kwargs: Any) -> "TileModule":
...
             if device.type == "cuda":
                 self.cuda(device)
             else:
                 self.cpu()
 
         if len(new_args) > 0 or len(kwargs) > 0:
             self._apply_without_context(lambda t: t.to(*new_args, **kwargs))
+
+        if dtype is not None:
+            # for the torch tile. This renews the hook with the correct dtype
```
maljoras, line 247: Should also be the case for the other tiles right? Best delete comment here. Don't specialize generic function for a particular tile.

tests/test_torch_tiles.py
```diff
@@ -476,3 +489,109 @@ def set_bm_nm(
...
+    linear_torch = linear_torch.to(device=device_type, dtype=dtype)
+    inp = randn((10, 256)).to(device=device_type, dtype=dtype)
+    out: Tensor
+    out = linear_torch(inp)  # pylint: disable=not-callable
+    assert out.dtype == dtype
+
+
+@mark.parametrize("dtype", [float32, float16, bfloat16])
+@mark.parametrize("device_type", [device("cpu"), device("cuda")])
+@mark.parametrize("use_autocast", [True, False])
```
maljoras, line 525: While these tests are nice, they fall out of the usual practice in the aihwkit. We actually have a test suite and parameterized test cases in tests/helpers that should be used an extended for uniformity in the aihwkit.

tests/test_utils.py
```diff
@@ -279,13 +267,9 @@ def test_save_load_state_dict_train_after_old_model(self):
...
-        (
-            new_model_weights,
-            new_model_biases,
-            new_tile_weights,
-            new_tile_biases,
-            _,
-        ) = self.get_layer_and_tile_weights(model)
+        (new_model_weights, new_model_biases, new_tile_weights, new_tile_biases, _) = (
+            self.get_layer_and_tile_weights(model)
+        )
```
maljoras, line 272: Do you use a different version of black so that you get this formatting changes? There is a specific version set in the requirements. please update that or use the same version

maljoras (changes requested):

src/aihwkit/simulator/tiles/module.py
jubueche: Yes, the `.get` was wrong. Don't know why this wasn't picked up by the tests.
I think the new logic handles all the cases correctly when we pass the `device` and `dtype` as either args and kwargs. The old logic didn't handle that correctly.

src/aihwkit/simulator/tiles/module.py
jubueche: Removed that.

tests/test_utils.py
jubueche: Yes I was using a newer version. I changed it back.

tests/test_torch_tiles.py
jubueche: For the torch tile, we have some other tests that are in this format because I wanted to test specific parameters. The other option would be to add these parameters as well before the `class TorchInferenceTest(ParametrizedTestCase):`, right?

coreylammie (approved):

Issue #287, opened 2021-09-15
synaptic bit-slicing


saba-er:
Hi,
I was wondering if and how synaptic bit-slicing is supported by the simulator? For example, the needed weight precision is 4 bits but the device only supports 2 bits and hence you need to use 2 devices per synaptic weight. Then, the analog computation (e.g. multiplication) will take into account the fact that the two devices combined represent a single weight (might be partially handled in the digital periphery?).

HeatPhoenix:
> Hi @SaBaKa2020 , thanks again! Currently, the simulator does not do bit-slicing in the way you are describing it by default, however, this is definitely on our roadmap to add tools to support that in a convenient way. With the `VectorUnitCell` multiple devices per crosspoint can be used, as well as a significance factor, however, the currents would always be added up "in analog" i.e. before the ADC.

@maljoras, can a `VectorUnitCell` setup be combined with, say, `InferenceRPUConfig` in the sense of rpu_config contains a vector of devices -> with each device is (inference only w/ a given noise model) contributing a part of the full weight. In other words, can we achieve this bit-slicing within an inference only setup? 

I could not find an example describing such a usage, only this [example](https://github.com/IBM/aihwkit/blob/d993b5ae50ac4d61dec0854de66812a439bc987b/examples/07_simple_layer_with_other_devices.py#L30).

HeatPhoenix:
Opened a pull request in #631 to hopefully help implement this and solve this issue.

maljoras-sony:
`VectorUnitCell` is more meant to explore multiple devices during in-situ training, that are also combine currents before the ADC conversion.   The example I gave above using `AnalogLinaer` would do the gradient through the bit-sliced layers. However, if the ConductanceConverter is used (where one could also use multiple weight slices), then the slicing is done after convergence (that is training the DNN without the bit-sliced weights). So these are all valid approaches, but slightly different concepts/

PR #631, opened 2024-03-27
Implemented linear synaptic bit-slicing layer

HeatPhoenix:
## Related issues

Pull request opened to contribute to Issue #287. 

## Description

I felt that bit-slicing may give a more realistic approximation on how current crossbar arrays are implemented in hardware (as to represent 8 or 16 bits, multiple devices are often used), and found issue #287 with @maljoras's instructions on how to implement this sufficient to try contributing a solution. Main difference from this implementation is that the number of slices can be arbitrarily set by the user.

This pull request includes mainly `linear_sliced.py`, which contains the implementation of the new module. The implementation is fairly naive, and is likely lacking some details that may stand out to someone more experienced with the codebase. I have tested its functionality by comparing its performance to a regular `AnalogLinear` using the `simple_layer.py` example.

I would highly appreciate feedback on the implementation.

## Details

`linear_sliced.py` implements `AnalogLinearBitSlicingLayer`. The constructor adds the following parameters to `AnalogLinear`:
`number_slices`, which sets the number of slices the weights should be divided over. `evenly_sliced`: whether the slices should have equal value (factor of 1) when `True`, or should be generated as MSB to LSB (factor of 2^x) when `False`. Finally, `significance_factors`, which allows for arbitrary setting of the factors, it must have a length equal to `number_slices`.

The `to_digital` (recombine all weights/factors together and assign) and `to_analog` (take digital weights and divide/split over number of slices) methods are also implemented, but since `Linear` would map to `AnalogLinear` I'm not sure how necessary or useful this is. 
Specifically for `to_analog`, I did not see an elegant way to let the user set how many slices they'd like to generate or other parameters. As stated previously, feedback is highly appreciated, I'd like this contribution to both be useful and properly implemented.

### DIFF
--- a/src/aihwkit/nn/modules/linear_sliced.py
+++ b/src/aihwkit/nn/modules/linear_sliced.py
@@ -22,10 +22,11 @@ from aihwkit.simulator.parameters.base import RPUConfigBase
 from aihwkit.nn import AnalogLinear
 
 class AnalogLinearBitSlicing(AnalogLayerBase, Linear):
-    """Linear layer that uses an analog tile.
+    """Linear layer that implements bit-slicing in an analog tile.
 
-    Linear layer that uses an analog tile during its forward, backward and
-    update passes.
+    A linear layer that implements bit-slicing in an analog tile during its forward, backward and
+    update passes. The bit-slicing is implemented by adding multiple AnalogLinear layers over which
+    the weights are sliced and distributed
 
     Note:
         The tensor parameters of this layer (``.weight`` and ``.bias``) are not
@@ -38,11 +39,16 @@ class AnalogLinearBitSlicing(AnalogLayerBase, Linear):
     Args:
         in_features: input vector size (number of columns).
         out_features: output vector size (number of rows).
+        number_slices: number of slices the weights should be distributed over
         bias: whether to use a bias row on the analog tile or not.
             for setting initial weights and during reading of the weights.
+        evenly_sliced: whether to slice the weights evenly over all slices, or to
+            go from MSB to LSB (factor = 2^x over the slices)
         rpu_config: resistive processing unit configuration.
         tile_module_class: Class for the tile module (default
             will be specified from the ``RPUConfig``).
+        significance_factors: the factors by which each slice is multiplied, can
+            be specified per slice
     """
 
     # pylint: disable=abstract-method
@@ -140,17 +146,22 @@ class AnalogLinearBitSlicing(AnalogLayerBase, Linear):
                 :class:`~aihwkit.simulator.tiles.array.TileModuleArray`
 
         Returns:
-            an AnalogLinearBitSlicing layer based on the digital Linear ``module``. Defaults to evenly sliced weights.
+            an AnalogLinearBitSlicing layer based on the digital Linear ``module``. Defaults to evenly sliced weights, with 8 slices.
         """
         analog_layer = cls(
             module.in_features,
             module.out_features,
+            8,
             module.bias is not None,
+            True,
             rpu_config,
             tile_module_class,
         )
-        
-        analog_layer.set_weights(module.weight, module.bias)
+
+        #slice total weight over number of slices and distribute over each slice part of the weight evenly
+        for idx, weight_slice in enumerate(analog_layer.analog_slices):
+            analog_layer.analog_slices[idx].set_weights(module.weight/len(analog_layer.analog_slices), module.bias/len(analog_layer.analog_slices))
+
         return analog_layer.to(module.weight.device)
 
     @classmethod
@@ -168,7 +179,8 @@ class AnalogLinearBitSlicing(AnalogLayerBase, Linear):
             as the analog linear layer.
         """
         weight, bias = 0, 0
-        #loop over slices, multiply factors with weights, add and give as weight/bias
+
+        #loop over slices, multiply factors with weights, add and give as weight/bias -- recombining back to original weight
         for idx, weight_slice in enumerate(self.analog_slices):
             slice_weight, slice_bias = weight_slice.get_weights(realistic=realistic)
             weight += slice_weight*self.significance_factors[idx]
--- a/src/aihwkit/nn/modules/linear_sliced.py
+++ b/src/aihwkit/nn/modules/linear_sliced.py
@@ -1,0 +1,181 @@
+# -*- coding: utf-8 -*-
+
+# (C) Copyright 2020, 2021, 2022, 2023, 2024 IBM. All Rights Reserved.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Analog layers."""
+from typing import Optional, Type, List
+
+from torch import Tensor
+from torch.nn import Linear, ParameterList
+
+from aihwkit.exceptions import ModuleError
+from aihwkit.nn.modules.base import AnalogLayerBase
+from aihwkit.simulator.parameters.base import RPUConfigBase
+from aihwkit.nn import AnalogLinear
+
+class AnalogLinearBitSlicing(AnalogLayerBase, Linear):
+    """Linear layer that uses an analog tile.
+
+    Linear layer that uses an analog tile during its forward, backward and
+    update passes.
+
+    Note:
+        The tensor parameters of this layer (``.weight`` and ``.bias``) are not
+        guaranteed to contain the same values as the internal weights and biases
+        stored in the analog tile. Please use ``set_weights`` and
+        ``get_weights`` when attempting to read or modify the weight/bias. This
+        read/write process can simulate the (noisy and inexact) analog writing
+        and reading of the resistive elements.
+
+    Args:
+        in_features: input vector size (number of columns).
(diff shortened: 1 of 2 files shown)


maljoras:
@HeatPhoenix very nice contribution, many thanks.

src/aihwkit/nn/modules/linear_sliced.py
```diff
@@ -0,0 +1,197 @@
...
+        """Compute the forward pass."""
+        # pylint: disable=arguments-differ, arguments-renamed
+        forward_output = 0 #to initialize the forward_output for the loop
+        for idx, weight_slice in enumerate(self.analog_slices):
+            forward_output += (weight_slice(x_input) * self.significance_factors[idx]) 
+
+        return forward_output  # type: ignore
+
+    @classmethod
+    def from_digital(
```
maljoras, line 132: It would be better to add some arguments in this from_digital call to set the slices etc. To use the convert_to_digital one has to use a custom mapping anyway. So the best would be to add a `convert_to_digital_sliced` function that converts all `Linear` to AnalogLinearSliced` which also accepts and passes the slicing parameters during the construction.

src/aihwkit/nn/modules/linear_sliced.py
```diff
@@ -0,0 +1,197 @@
...
+from torch import Tensor
+from torch.nn import Linear, ParameterList
+
+from aihwkit.exceptions import ModuleError
+from aihwkit.nn.modules.base import AnalogLayerBase
+from aihwkit.simulator.parameters.base import RPUConfigBase
+from aihwkit.nn import AnalogLinear
+
+
+class AnalogLinearBitSlicing(AnalogLayerBase, Linear):
```
maljoras, line 25: Actually it would be better if you would just write a new module which inherits from `AnalogContainerBase` or simply from `Module` instead of `AnalogLayerBase`. This is because if `AnalogLinear` layers are used as children of an `AnalogLayer` which will cause issues. See for instance the RNN module that internally also use `AnalogLinear` layers,  see [here](https://github.com/IBM/aihwkit/blob/69e524fbc2c5462cbe1a12ee1662879c001c30c0/src/aihwkit/nn/modules/rnn/layers.py#L20) and  [here](https://github.com/IBM/aihwkit/blob/69e524fbc2c5462cbe1a12ee1662879c001c30c0/src/aihwkit/nn/modules/rnn/rnn.py#L128).

maljoras: It might also not necessary to inherit from `Linear` at all, as this will cause additional issues with the `Linear` weight etc.

maljoras (changes requested):

maljoras:
Many thanks @HeatPhoenix for this great contribution. I think it needs some adaption, see my comments above. Also would be great if you would add some unit tests and an example.

src/aihwkit/nn/modules/linear_sliced.py
HeatPhoenix: convert_to_analog_sliced, you mean?

kaoutar55:
@charles-mackin can you please review this PR and check how it is aligned with your implementation of bit-slicing

HeatPhoenix:
FWIW, I still intend to follow up on the comments by @maljoras, just haven't had time to as of yet.

charlesmackin:
It seems weight bit slicing can be implemented (within a unit cell) by simply writing a new g_converter class as seen here: https://github.com/IBM/aihwkit/blob/8cb5063eb6dc1d352e90d0786c45109de4f4ac13/src/aihwkit/inference/converter/conductance.py#L218.

The CustomPairConductanceConverter class already almost does this - just in a non-quantized (interpolated for continuous weights) way. The f_lst parameter can be used to set the significance to equivalent 1s, 2^N, or arbitrary values. I would suggest creating a new class and modifying the convert_to_conductance method to quantize the weights as opposed to interpolating (as is done currently) and a new convert_back_to_weights method to accomplish this. It seems a much simpler way to implement weight bit slicing. It should also be compatible with linear layers, convs, etc.  It should also be compatible with activation slicing (e.g. bit-wise and split-PWM modes).

For example, you should be able to simply provide parameters below for +/- weight programming for a 4-device unit cell with 2^N weighting.

``` 
f_lst = [1, 2]
g_lst = [[0, 0, 0, 0, 1, 0, 1], # gp0
              [1, 0, 1, 0, 0, 0, 0], # gm0
              [0, 0, 0, 0, 0, 1, 1], # gp1
              [1, 1, 0, 0, 0, 0, 0]] # gm1
#            -1 -0.67 -0.3 0 0.3 0.67 1  in unitless weights
```
# replace 0, 1 with g_min, g_max to specify how to program device

Use the modified convert_to_conductance method to convert continuous weight distributions to a quantized one using a scheme similar to the one above.

I think you could even substitute with non g_min, g_max values if you want to enable less trivial programming strategies. For instance, programming at the absolute g_max of a device may have some undesirable characteristics. You could easily substitute 1 = 0.9 * g_max which may program better at a slightly reduced weight range. Same goes with g_min.

HeatPhoenix:
> It seems weight bit slicing can be implemented (within a unit cell) by simply writing a new g_converter class as seen here:
> 
> https://github.com/IBM/aihwkit/blob/8cb5063eb6dc1d352e90d0786c45109de4f4ac13/src/aihwkit/inference/converter/conductance.py#L218
> 
> .
> The CustomPairConductanceConverter class already almost does this - just in a non-quantized (interpolated for continuous weights) way. The f_lst parameter can be used to set the significance to equivalent 1s, 2^N, or arbitrary values. I would suggest creating a new class and modifying the convert_to_conductance method to quantize the weights as opposed to interpolating (as is done currently) and a new convert_back_to_weights method to accomplish this. It seems a much simpler way to implement weight bit slicing. It should also be compatible with linear layers, convs, etc. It should also be compatible with activation slicing (e.g. bit-wise and split-PWM modes).
> 
> For example, you should be able to simply provide parameters below for +/- weight programming for a 4-device unit cell with 2^N weighting.
> 
> ```
> f_lst = [1, 2]
> g_lst = [[0, 0, 0, 0, 1, 0, 1], # gp0
>               [1, 0, 1, 0, 0, 0, 0], # gm0
>               [0, 0, 0, 0, 0, 1, 1], # gp1
>               [1, 1, 0, 0, 0, 0, 0]] # gm1
> #            -1 -0.67 -0.3 0 0.3 0.67 1  in unitless weights
> # replace 0, 1 with g_min, g_max to specify how to program device
> ```
> 
> Use the modified convert_to_conductance method to convert continuous weight distributions to a quantized one using a scheme similar to the one above.
> 
> I think you could even substitute with non g_min, g_max values if you want to enable less trivial programming strategies. For instance, programming at the absolute g_max of a device may have some undesirable characteristics. You could easily substitute 1 = 0.9 * g_max which may program better at a slightly reduced weight range. Same goes with g_min.

I'm not really deep into it enough anymore to understand if this implies what I've implemented is superfluous and should be re-implemented basically entirely differently?

charlesmackin:
@HeatPhoenix Yes, I believe it would be more appropriate (and likely simpler) to implement the synaptic bit slicing via a new ConductanceConverter class. It would be great if @maljoras could provide his thoughts as well here.

src/aihwkit/nn/modules/linear_sliced.py
maljoras-sony: Right

maljoras-sony:
Hi @charlesmackin,  `g_converter` indeed could handle one form of bit-slicing and it makes sense in some use cases. However, it would only slice the weights `after` the convergence of the DNN so there is no gradient pass through the bit-sliced layers. Also, the bit-slices are summing currents in analog and then are subject to digital conversion. On the other hand, the approach by @HeatPhoenix I think was to use AnalogLinear as one of the slices, which would mean that the gradient could pass through the different slices during training and also that the summing of the outputs is done in digital instead. So there are slight differences to the approaches.

HeatPhoenix:
> Hi @charlesmackin, `g_converter` indeed could handle one form of bit-slicing and it makes sense in some use cases. However, it would only slice the weights `after` the convergence of the DNN so there is no gradient pass through the bit-slices of the weights. Also, the weight-slices are summing currents in analog and only then are subject to digital conversion. On the other hand, the approach by @HeatPhoenix I think was to use `AnalogLinear` as the base of the weight slices, which would mean that the gradient could pass through the individual slices during training and also that the summing of the outputs is done in digital instead. So there are slight differences to the approaches.

In this case, should I move forward with the improvements as outlined by @maljoras-sony previously? As an alternative to @charlesmackin's implementation.

charlesmackin:
@HeatPhoenix Yes, you should move forward with @maljoras-sony suggestions in this case

PabloCarmona:
@HeatPhoenix did you had the chance to take a look at @maljoras-sony suggestions like @charlesmackin advice you? Let us know if you need any further help to be able to finish this PR. Thanks!

HeatPhoenix:
> @HeatPhoenix did you had the chance to take a look at @maljoras-sony suggestions like @charlesmackin advice you? Let us know if you need any further help to be able to finish this PR. Thanks!

Still interested in finishing up, just really strapped for time, generally. Will try to get it done near the start of the new year.

kaoutar55:
Thanks @HeatPhoenix please let us know when you have any new updates!

PabloCarmona:
Hello @HeatPhoenix! Did you have any updates on this? Can we help you out in some manner? For us this is a good addition to the kit capabilities and we want to add it, thanks for your work!

kaoutar55:
@HeatPhoenix are you still interested in continuing to work on this issue. Please let us know. Thanks!

PR #638, opened 2024-04-05
Update README.md

samsung-ro:


maljoras:
@kaoutar55 Is this something from you? Looks almost like spam. There should be description of the patch and the motivation always. Why would we change the license?

kaoutar55:
This was not from me and we should discard this.

PR #648, opened 2024-05-01
Added Example 30

coreylammie:
This example demonstrates how to import and perform inference using a model which has been trained in a hardware-aware fashion using an external library (i.e., not the AIHWKIT).

The external model is in the form of a standard pytorch model with hardware-aware trained weights. Input and output bounds, in addition to output scales are not defined. Here, we set these using input range calibration and determine the model accuracy at different evaluation (drift) times.

### DIFF
--- a/examples/30_external_hardware_aware_model.py
+++ b/examples/30_external_hardware_aware_model.py
@@ -1,0 +1,316 @@
+# type: ignore
+# pylint: disable=missing-module-docstring
+# pylint: disable=missing-class-docstring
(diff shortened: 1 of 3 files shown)


coreylammie:
_**Note:**_ Before this is merged, the checkpoint file `resnet9s.th` needs to be uploaded to the tutorial COS and downloaded in the example using wget or a python-equivalent library.

jubueche (dismissed):

jubueche (approved):

PR #674, opened 2024-08-07
Updated drift compensation mechanism to use `is_perfect=True` when determining the reference readout

coreylammie:
## Description

Updated drift compensation mechanism to use `is_perfect=True` when determining the reference readout. Previously, all non-idealities (as set by the corresponding `rpu_config`) were modeled, resulting in sub-optimal drift compensation scales being computed in some scenarios, e.g., where the output noise is sufficiently large.

### DIFF
--- a/src/aihwkit/simulator/tiles/inference.py
+++ b/src/aihwkit/simulator/tiles/inference.py
@@ -98,7 +98,7 @@ class InferenceTileWithPeriphery(TileWithPeriphery):
             self.set_mapping_scales(mapping_scales)
 
     @no_grad()
-    def _forward_drift_readout_tensor(self, reset_if: bool = False) -> Optional[Tensor]:
+    def _forward_drift_readout_tensor(self, reset_if: bool = False, is_perfect=False) -> Optional[Tensor]:
         """Perform a forward pass using the drift read-out tensor.
 
         Args:
@@ -126,9 +126,16 @@ class InferenceTileWithPeriphery(TileWithPeriphery):
 
         # We need to take the bias as a common column here, also we do
         # not want to use indexed.
-        return self.tile.forward(
+        is_perfect_state = self.rpu_config.forward.is_perfect
+        tile_rpu_config = self.rpu_config
+        tile_rpu_config.forward.is_perfect = True
+        self.tile.set_config(tile_rpu_config)
+        output = self.tile.forward(
             self.drift_readout_tensor, False, self.in_trans, self.out_trans, True, self.non_blocking
         )
+        tile_rpu_config.forward.is_perfect = is_perfect_state
+        self.tile.set_config(tile_rpu_config)
+        return output
 
     @no_grad()
     def program_weights(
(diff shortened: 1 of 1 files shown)


maljoras:
@coreylammie, the drift compensation is on purpose using the rpu config of the tile as this is the realistic situation. In really how can you achieve a perfect reference? I don't think you should change this behaviour as it would then make the simulation unrealistic and wrong.

maljoras:
If you want perfect readout during drift compensation, then you should explicitly write a new drift compensation. This should be a special case as this behaviour is wrong in most natural cases.

maljoras:
This PR will also invalidate the papers we have written using this standard drift where the correct way was used to estimate the drift compensation in a noisy manner. So please take this PR back. If you need a perfect compensation for some reasons, write a special drift compensation class by deriving from the default one.

maljoras:
Also the implementation seems to permanently set the rpu config to perfect, which is then always using the perfect case, which is even wrong, if you only want to use a noise free reference. It will then use a completely noise free drift. That's completely wrong.

maljoras:
Please don't change the default drift class in this manner. Use a special drift class if you need this special behaviour for some reasons.

maljoras:
I am about to close this PR. Please submit a new one where you modify your PR accordingly by deriving a new drift compensation class from the default class and not changing the behaviour of the default class. Also, as stated above, it seems that the rpu configuration is permanently changed which is wrong in any case.

maljoras:
Btw, if you want the exact weights of a tile for the reference computation of your special drift compensation class, simply use `tile.get_weights(realistic=False)` instead of changing the forward read to with `is_perfect=True`

kaoutar55:
@coreylammie can you please let us know what is the latest on this PR. Should we go with the changes you have? We need to close this PR soon.

src/aihwkit/simulator/tiles/inference.py
```diff
@@ -170,16 +205,17 @@ def program_weights(
...
 
         self.tile.set_weights(self.programmed_weights)
 
         if (
             hasattr(self.rpu_config, "drift_compensation")
             and self.rpu_config.drift_compensation is not None
         ):
-            forward_output = self._forward_drift_readout_tensor(True)
-            self.drift_baseline = self.rpu_config.drift_compensation.init_baseline(forward_output)
+            self.drift_baseline = self.rpu_config.drift_compensation.init_baseline(self)
```
maljoras-sony, line 218: I think here is still a conceptual error as far as I can see. If you simply set `is_perfect=True` then the `programmed_weights` are used to establish the reference without any additional noise. However, this cannot be done in reality as the programmed weights are those in the conductances. What you want to do is taking the self.reference_combined_weights as the reference for the drift, since those weights are the floating point weights that are then programmed onto the crossbar. This is a difference, because the programmed weights night in fact have some corrupt devices etc. and thus different from the floating point weights even if `is_perfect=True` is set.

maljoras-sony (changes requested):

src/aihwkit/simulator/tiles/inference.py
coreylammie: Makes sense - thanks for the explanation! I'll change this now. As an aside, I will raise a separate issue to add a warning/note to the documentation here: https://aihwkit.readthedocs.io/en/latest/api/aihwkit.simulator.parameters.io.html#aihwkit.simulator.parameters.io.IOParameters.is_perfect. It is somewhat ambiguous what a forward pass is and whether a user would expect the FP or programmed weights to be used to compute it.

coreylammie: @maljoras could you please check my recent code changes? I have used `self.reference_combined_weights`, as suggested.

kaoutar55:
@maljoras-sony @maljoras can you please review the changes that Corey did to address your review.

kaoutar55:
Hi @maljoras-sony can you please check if you agree with these changes so we can close this PR?

PR #685, opened 2024-08-22
add release content for 0.9.2

PabloCarmona:
PR for release content of aihwkit v0.9.2:

* New Hermes noise model
* New conductance converters
* Make conv also compatible with non-batched input
* Add per column drift compensation
* Update requirements-examples.txt
* Add custom drift


src/aihwkit/inference/noise/pcm.py
```diff
@@ -34,7 +39,7 @@ class PCMLikeNoiseModel(BaseNoiseModel):
     r"""Noise model that was fitted and characterized on real PCM devices.
 
     Expected weight noise at assumed time of inference with expected
-    programming noise at 0.
+/dccstor/transformer/charles/aihwkit/src/aihwkit/utils    programming noise at 0.
```
jubueche, line 42: This seems like it shouldn't be there.

jubueche: Also, the new examples/notebooks need to be run at least one time before merging this. @PabloCarmona can you please do this? When they are working, you can merge.

jubueche: Also, `FAILED tests/test_simulator_tiles.py::TileForwardBackwardTest_ConstantStep::test_set_forward_out_noise_std` failed. That also needs to pass. Maybe run it again and see if it was just a rounding error.

PabloCarmona: I passed that test three times and OK for me. Also I'm gonna remove that text you pointed out and review the notebook. Thanks!

PabloCarmona: Confirmed that in CUDA enabled envs is passing also.

kaoutar55 (commented):
can you add the changes to the changelog in this PR and also the changes to the version file.

examples/31_custom_drift_models.py
```diff
@@ -0,0 +1,151 @@
...
+
+# define rpu_config
+rpu_config = TorchInferenceRPUConfig()
+rpu_config.noise_model = PCMLikeNoiseModel(
+    prog_noise_scale=0.0,   # turn off to show drift only
+    read_noise_scale=0.0,   # turn off to show drift only
+    drift_scale=1.0,
+    g_converter=SinglePairConductanceConverter(g_min=g_min, g_max=g_max),
+    custom_drift_model=custom_drift_model)
+rpu_config.forward.bound_management = BoundManagementType.NONE
```
maljoras, line 69: It would be better to use the dataclass instantiation way to set the parameter in the example, like: 
```
io_pars = IOParameter(nm_thres=1.0, inp_res=2 **8 -2 ,....)
noise_model = PCMLikeNoiseModel(prog_noise_scale=0.0, ....)
rpu_config = TorchInferenceRPUConfig(noise_model=noise_model, forward=io_pars)
```
This would avoid the case that accidentally wrong assignments are made. Maybe better show-case this "more correct" way of building the `rpu_config`

src/aihwkit/inference/noise/hermes.py
```diff
@@ -0,0 +1,277 @@
...
+        g_max: Optional[float] = None,
+        t_read: float = 512.0e-9,
+        t_0: Optional[float] = None,
+        prog_noise_scale: float = 1.0,
+        read_noise_scale: float = 1.0,
+        drift_scale: float = 1.0,
+        prog_coeff_g_max_reference: Optional[float] = None,
+    ):
+        # The only valid options are 1 or 2 devices
+        assert num_devices in [1, 2], "Hermes supports either 1 or 2 devices per weight"
```
maljoras, line 105: In theory the number of devices should be set by the `g_converter`.

src/aihwkit/inference/noise/pcm.py
```diff
@@ -102,6 +108,34 @@ def __init__(
         self.prog_noise_scale = prog_noise_scale
         self.read_noise_scale = read_noise_scale
         self.drift_scale = drift_scale
+        self.valid_kwargs = ['custom_drift_model']
+
+        if not all(key in self.valid_kwargs for key in kwargs):
+            ValueError("PCMLikeNoiseModel only supports kwargs = %s" % self.valid_kwargs)
+
+        self.custom_drift_model = kwargs.get('custom_drift_model')
```
maljoras, line 116: These "hidden" args are not very user friendly. Why not simply derive a new class from the PCMLikeNoiseModel that has a custom drift model? Then one does not get into problems with the checkpoint read of earlier files etc.

src/aihwkit/inference/noise/pcm.py
```diff
@@ -102,6 +108,34 @@ def __init__(
...
+
+        if self.custom_drift_model is not None:
+            drift_model_g_min = min(self.custom_drift_model['g_lst'])
```
maljoras, line 119: These hidden args are bad code practice. Why not give these explicitly in a new class ?

src/aihwkit/inference/noise/pcm.py
```diff
@@ -78,6 +83,7 @@ def __init__(
         read_noise_scale: float = 1.0,
         drift_scale: float = 1.0,
         prog_coeff_g_max_reference: Optional[float] = None,
+        **kwargs: Any,
```
maljoras, line 86: Best not to use "kwargs" here. Why not simply add new explicit args? If this changes old checkpoints, maybe best to just write a new noise model class with custom drift model?

src/aihwkit/inference/noise/pcm.py
```diff
@@ -125,13 +159,62 @@ def apply_programming_noise_to_conductance(self, g_target: Tensor) -> Tensor:
...
+        g_lst, a list of conductances in ascending order; nu_mean_lst, a list
+        of mean drift coefficients corresponding to the g_lst values: and
+        nu_std_lst, a list of nu standard deviation values corresponding to
+        the g_lst values. Nu coeffiecients will be interpolated using this
+        model information."""
+
+        if self.custom_drift_model is not None:
+            assert isinstance(self.custom_drift_model,
+                              dict), "custom_drift_model must be specified as dictionary"
+            required_keys = ['g_lst', 'nu_mean_lst', 'nu_std_lst']
```
maljoras, line 173: see above.

src/aihwkit/nn/modules/conv.py
```diff
@@ -180,12 +180,14 @@ def forward(self, x_input: Tensor) -> Tensor:
             dilation=self.dilation,
             padding=self.padding,
             stride=self.stride,
-        ).transpose(1, 2)
+        ).transpose(-1, -2)
 
-        out = self.analog_module(x_input_).transpose(1, 2)
+        out = self.analog_module(x_input_).transpose(-1, -2)
```
maljoras, line 185: I hope this was tested in all possible input cases (in_trans, out_trans) and all tiles + indexed and non-indexed convolution? It does look rather wrong to me to use -1 -2 instead of 1, 2

maljoras (changes requested):

PabloCarmona:
Thanks @maljoras for the review, we are going to take actions and see it and work through your suggestions.

PabloCarmona:
> can you add the changes to the changelog in this PR and also the changes to the version file.

Done!

examples/31_custom_drift_models.py
PabloCarmona: Done!

src/aihwkit/inference/noise/hermes.py
Borjagodoy: I see here a problem @maljoras Feel like the g_converter is a optional parameter and the num_devices not, what's happened when no exist the g_converter? i think that should be by default 2

src/aihwkit/nn/modules/conv.py
PabloCarmona: I run the tests with the change and everything seems to pass and look fine. Also, @jubueche runs the tests, and they work well for him.

src/aihwkit/inference/noise/hermes.py
atvasilopoulos: In the Hermes model, the number of devices has the effects of 
- changing the programming method from normal iterative programming to MSF
- changing the drift model variability based on on-chip measurements.

As such, the g_converter number of devices argument cannot be used for this functionality, leading to a discrete argument. I added some additional details on what the selection of `num_devices` incurs in the docstring of the function (see #691)

CHANGELOG.md
```diff
@@ -12,6 +12,20 @@ The format is based on [Keep a Changelog], and this project adheres to
 * `Fixed` for any bug fixes.
 * `Security` in case of vulnerabilities.
 
+## [0.9.2] - 2024/08/26
```
kaoutar55, line 15: Please put the related PRs in the changelog similar to how we did it in the past.

kaoutar55 (commented):
Please add the issue numbers in the change log for each line when applicable

CHANGELOG.md
PabloCarmona: Done!

kaoutar55 (approved):

Borjagodoy (approved):

PR #696, opened 2024-10-22
License changes

PabloCarmona:
## Related issues

License change from Apache 2.0 to MIT.


jubueche:
@PabloCarmona why are all the tests not passing? And why are there so many other commits? For this change, I am just expecting a change that changes the license.

Issue #699, opened 2024-12-04
Output image quality issues in RGB image generation tasks


sheep1222:
I want to use this tool to simulate, train, and infer with a convolutional neural network for RGB image generation tasks, which incorporates an attention mechanism. I have tried using the convert_to_nalog function and various RPU configurations, but the results were not satisfactory. So, I would like to ask, which configuration you would recommend without considering the non-idealities of the devices, or is there a method to directly map the trained parameters for inference?

PabloCarmona:
Hello @sheep1222, and thanks for trying the kit! Can you please elaborate more on your setup with some code examples of what you are trying to achieve?

sheep1222:
Sorry, I've been busy lately and haven't replied in a timely manner.
My program is like this：
import torch
import torch.nn as nn
import torchvision.utils as vutils
import torch.optim as optim
import torchvision
from torchvision import datasets, transforms
from torch.utils.data import DataLoader
import matplotlib.pyplot as plt
import numpy as np
import imageio
from aihwkit.nn import AnalogLinear, AnalogSequential
from aihwkit.optim import AnalogSGD
from aihwkit.nn.conversion import convert_to_analog
from aihwkit.simulator.presets import MixedPrecisionEcRamMOPreset
from aihwkit.simulator.rpu_base import cuda
from aihwkit.simulator.configs import MappingParameter

mapping = MappingParameter(weight_scaling_omega=0.8)
RPU_CONFIG = MixedPrecisionEcRamMOPreset(mapping=mapping)

# Hyperparameters
batch_size = 128
image_size = 64
nc = 3
nz = 100
ngf = 64
ndf = 64
num_epochs = 5
lr = 0.0002
beta1 = 0.5

USE_CUDA = 0
if cuda.is_compiled():
    USE_CUDA = 1
device = torch.device("cuda" if USE_CUDA else "cpu")

# Data Preparation
transform = transforms.Compose([
    transforms.Resize(image_size),
    transforms.CenterCrop(image_size),
    transforms.ToTensor(),
    transforms.Normalize([0.5] * nc, [0.5] * nc)
])

dataset = datasets.ImageFolder(root="./data/", transform=transform)
dataloader = DataLoader(dataset, batch_size=batch_size, shuffle=True, num_workers=4)


# Generator
class Generator(nn.Module):
    def __init__(self):
        super(Generator, self).__init__()
        self.main = nn.Sequential(
[...]

PR #705, opened 2025-01-06
added FP preset

jubueche:
## Related issues

https://github.com/IBM/aihwkit/issues/699

## Description

Add a preset that can be used to do FP inference and training.

## Details

- New preset for FP training/inference
- Added test in preset tests to check correctness on tile level


PabloCarmona:
Seems like there are some technical problems on the side of Travis CI and there on the way to fixing it. Give them a couple of days to solve it and we will try to trigger the build.

PabloCarmona (approved):
Looks good to me! If the build, compile and tests passes we can go ahead and merge! There was having problem with travis that seems to be solved and stable from yesterday night.

PR #732, opened 2025-06-11
Fix chopper bug

### LINKED ISSUE #642: AnalogConv2d fails when using TT-v2
## Description
When I tried to use the TT-v2 algorithm to train the convolutional network, I got a Cuda error.

## How to reproduce
After running the following `main.py` file, I got an error `RuntimeError: CUDA_CALL Error 'an illegal memory access was encountered' at cuda_util.cu:653`
```python
# main.py
import torch
from aihwkit.nn import AnalogConv2d
from aihwkit.optim import AnalogSGD
from aihwkit.simulator.configs import build_config
from aihwkit.simulator.configs.devices import SoftBoundsReferenceDevice
DEVICE = 'cuda:0'

rpu_config = build_config('ttv2', device=SoftBoundsReferenceDevice())
model = AnalogConv2d(
    in_channels=1, out_channels=3, kernel_size=5, rpu_config=rpu_config
).to(DEVICE)

optimizer = AnalogSGD(model.parameters(), lr=0.1)
optimizer.regroup_param_groups(model)
# if I use images = torch.empty((128, 1, 32, 32)), even the forward fails
images = torch.ones((128, 1, 32, 32))
images = images.to(DEVICE)
output = model(images)
loss = output.norm()**2
loss.backward()
optimizer.step()
```
Besides, if I create `torch.empty()` instead of `torch.one()`, the forward clause `model(images)` never stops, I guess there could be some endless loop happening.

## Other information

* Pytorch version: 2.1.2
* Package version: aihwkit-gpu 0.9.0
* OS: Linux
* Python version: 3.10.13
* Conda version (or N/A) :23.11.0

maljoras-sony:
## Related issues

Fixes #642 and other issues.

## Description

Critical update for a number of issues in the C++/CUDA backend for analog training related devices (TTv2 / TTv3 / AGAD). The maximizer bug could have affected other CUDA runs as well (transposed forward / backward / update).  Now all C++ tests are running, including #642. As a result, now the max memory usage is greatly reduced for analog training, making it possible to run larger models and batch sizes.

## Details
A number of bugs that were fixes in this PR:
* Cmake issues for debug mode and C++ test generation is fixed. Use e.g. `make flags="-DRPU_DEBUG=ON -DBUILD_TEST=ON -DRPU_USE_TORCH_BUFFERS=OFF" build_inplace_cuda`. Tests will remain in the build directory, e.g. `_skbuild/linux-x86_64-3.11/cmake-build` 
* There was an issue with the CUDA array generation in `RPUCudaChoppedTransferDevice` which ran out of the `int` scope and memory. Such a large array is not needed at all; now it is changed to do chunking instead (limit can be set in C++ with the new parameter: `transfer_max_vec_chunk_size` ). Also, size variables of `CudaArray` is now proper `size_t` 64-bit to avoid any kind of overflow. 
* If `flexible_insize` is false (for some obscure forward MVM settings), there was a bug in the chopped weight updater resulting in the wrong update for ttv2 (when using the faster CUDA code for ttv2 and not the buffer device) and ttv3 and AGAD.
* maximizer on CUDA had an error for MAX FP atomics which could be a critical issue for some configuration (e.g. transposed vector update).  Might have been silent for other configs.  
* Added a few more tests for IO iterators


maljoras:
@PabloCarmona would be great to add this to a new release as it is an important bug-fix.

PR #770, opened 2026-06-08
[ROCm] Add AMD GPU (HIP) support to the RPUCuda backend

jeffdaily:
## Summary

This adds an optional `USE_HIP` build of the RPUCuda backend so the analog tile simulator runs on AMD GPUs through ROCm, alongside the existing `USE_CUDA` path, which is left byte-identical. The build stays on the project's standalone scikit-build/CMake CUDA toolchain (PyTorch is used only for headers and linkage), so no torch `cpp_extension` hipify is involved.

Build it with `-DUSE_HIP=ON` (mutually exclusive with `USE_CUDA`) and select the GPU with `-DCMAKE_HIP_ARCHITECTURES=<arch>` (e.g. `gfx90a`, `gfx1100`). The ROCm/HIP build is documented in the developer install guide alongside the CUDA build.

## What is implemented

- **`src/rpucuda/cuda/cuda_to_hip.h`** (the single HIP-aware header, included once from `cuda_util.h` under `USE_HIP`): aliases the `cuda*` / `cublas*` / `curand*` spellings to `hip*` / hipBLAS / hipRAND so the CUDA-spelled device sources compile unchanged. `curandState`, `curandState_t` and `curandStateXORWOW` (one XORWOW type in CUDA) are all mapped to one hipRAND type, since hipRAND declares them as distinct structs and splitting them would drop a `CudaArray` instantiation.
- **`bit_line_maker.cu`** warp-size handling: its stochastic pulse train packs a 32-lane ballot into 32-bit words, so the on-device layout must be independent of the physical wavefront width. The compat header redefines the ballot/shuffle sync intrinsics as width-32 logical-warp operations (a wave64 CDNA device selects the calling lane's own 32-lane subgroup), so wave32 and wave64 emit the same stream the CUDA build produces.
- **CMake**: `option(USE_HIP)` plus a `dependencies_hip.cmake` mirroring the CUDA path (HIP language, `.cu` as `LANGUAGE HIP`, hipBLAS/hipRAND/hipCUB, arch from `CMAKE_HIP_ARCHITECTURES`, IPO off since HIP device linking does not finalize LTO). A new `RPU_CXX_STANDARD` cache variable (default 17) lets the standard track newer PyTorch headers.
- **c10 stream shim** in `rpu_base_tiles_cuda.cpp`: keyed on the torch hipify generation (not the ROCm version) via a CMake-detected `TORCH_HIPIFY_V2` define, since which `c10` stream symbol is public depends on the hipify generation.
- **Windows / amdclang++**: guard the MSVC-only `/O2` flag with `if(MSVC)`, and fix a `class` vs `struct` forward-declaration mismatch that mangles differently on the MSVC ABI.

## Validation

The analog tiles are exercised on device against the CPU tiles (tolerance-based), via the pytest tile/binding/torch/inference/layer suites:

| GPU | Arch | OS / ROCm |
| --- | --- | --- |
| Instinct MI250X | gfx90a (CDNA2, wave64) | Linux, ROCm 7.2.1 |
| Radeon Pro W7800 | gfx1100 (RDNA3, wave32) | Linux, ROCm 7.2.1 |
| Radeon RX 9070 XT | gfx1201 (RDNA4, wave32) | Windows, ROCm 7.14 |

`test_specific_tiles` (the bit-line-maker + pulsed-update path) passes 18/18 on each, and the tile/binding/torch/inference/layer suites pass (one order-sensitive stochastic weight-programming test passes in isolation). A fat binary for `gfx90a;gfx1100` emits both code objects for `bit_line_maker.cu`, confirming the ballot rewrite is wave-width agnostic.

The CUDA/NVIDIA path is unchanged and not affected by this build.

This work was authored with the assistance of Claude, an AI assistant.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,9 +29,8 @@ set(RPU_CXX_STANDARD "17" CACHE STRING "C++ standard used for all RPU targets")
 
 set(RPU_BLAS "OpenBLAS" CACHE STRING "BLAS backend of choice (OpenBLAS, MKL)")
 set(RPU_CUDA_ARCHITECTURES "75;80;89" CACHE STRING "Target CUDA architectures")
-if(NOT DEFINED CMAKE_HIP_ARCHITECTURES)
-  set(CMAKE_HIP_ARCHITECTURES "gfx90a" CACHE STRING "Target HIP (ROCm) architectures")
-endif()
+# CMAKE_HIP_ARCHITECTURES is left to enable_language(HIP) to auto-detect the host
+# GPU when not set explicitly via -DCMAKE_HIP_ARCHITECTURES.
 
 # Internal variables.
 set(CUDA_TARGET_PROPERTIES POSITION_INDEPENDENT_CODE ON
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ project(aihwkit C CXX)
 option(BUILD_TEST "Build C++ test binaries" OFF)
 option(BUILD_EXTENSION "Build additional C++ tools" OFF)
 option(USE_CUDA "Build with CUDA support" $ENV{USE_CUDA})
+option(USE_HIP "Build with ROCm/HIP support" $ENV{USE_HIP})
 
 # experimental precision flags
 option(RPU_USE_FP16 "EXPERIMENTAL: Build FP16 support (only available with CUDA)" OFF)
@@ -22,14 +23,24 @@ option(RPU_USE_FASTRAND "Use fastrand" OFF)
 option(RPU_USE_TORCH_BUFFERS "Use torch buffers for RPUCuda" ON)
 
 
+# The C++ standard tracks the linked PyTorch headers; newer Torch releases
+# require C++20. Default stays 17 to match older Torch.
+set(RPU_CXX_STANDARD "17" CACHE STRING "C++ standard used for all RPU targets")
+
 set(RPU_BLAS "OpenBLAS" CACHE STRING "BLAS backend of choice (OpenBLAS, MKL)")
 set(RPU_CUDA_ARCHITECTURES "75;80;89" CACHE STRING "Target CUDA architectures")
+if(NOT DEFINED CMAKE_HIP_ARCHITECTURES)
+  set(CMAKE_HIP_ARCHITECTURES "gfx90a" CACHE STRING "Target HIP (ROCm) architectures")
+endif()
 
 # Internal variables.
 set(CUDA_TARGET_PROPERTIES POSITION_INDEPENDENT_CODE ON
                            CUDA_RESOLVE_DEVICE_SYMBOLS ON
                            CUDA_SEPARABLE_COMPILATION ON
-                           CXX_STANDARD 17)
+                           CXX_STANDARD ${RPU_CXX_STANDARD})
+set(HIP_TARGET_PROPERTIES POSITION_INDEPENDENT_CODE ON
+                          INTERPROCEDURAL_OPTIMIZATION OFF
+                          CXX_STANDARD ${RPU_CXX_STANDARD})
 
 # Append the virtualenv library path to cmake.
 if(DEFINED ENV{VIRTUAL_ENV})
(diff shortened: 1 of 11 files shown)


PabloCarmona:
Hello @jeffdaily! Thanks a lot for this PR! We are looking into this and try to get access to an AMD GPU to test it out.

Please can you add your signature to the commits so the DCO check can pass? Thanks again!

jeffdaily:
Thanks @PabloCarmona! I've added my Signed-off-by to the commit, so the DCO check passes now.

PabloCarmona:
Hello @jeffdaily! Sorry for the late response, but we are trying to test this in an environment with an AMD GPU. Could you share with us some evidences and results on how this works and runs on an AMD GPU env to double check it? Thanks!

jeffdaily:
*Note: this reply was drafted by an AI assistant.*

Sure. This was built and tested on four AMD configurations: an Instinct MI250X (gfx90a) and a Radeon Pro W7800 (gfx1100) on Linux with ROCm 7.2.1, and a Radeon RX 9070 XT (gfx1201) plus a Radeon PRO V710 (gfx1101) on Windows with a ROCm 7.14 nightly toolchain. That covers both AMD wavefront widths (64 on the Instinct part, 32 on the Radeon parts), which matters because the pulsed-update bit-packing code is warp-size sensitive.

Results at the current branch tip, all with the HIP build (`USE_HIP=ON`, `python setup.py build_ext`):

- `tests/test_specific_tiles.py`: 18/18 passed. This is the important one for the port: it exercises the bit_line_maker pulsed-weight-update path, where the warp-size differences live, and it compares CPU tiles against GPU tiles numerically.
- `tests/test_simulator_tiles.py` + `tests/test_bindings_tiles.py`: 530 passed, 56 skipped, 1 failed. The one failure is `TileTest_Inference.test_program_weights` on the CPU tile (not the GPU path): a stochastic 5%-tolerance convergence test that is sensitive to global RNG ordering. It passes 15/15 when run in isolation, and the GPU variant of the same test passes 8/8.
- `tests/test_torch_tiles.py` + `tests/test_inference_tiles.py`: 406 passed, 55 skipped.

The CUDA path was also rebuilt from the same branch with nvcc 12.8 (`USE_CUDA=ON`, sm_75/80/89) to confirm it still compiles cleanly, since the HIP support is additive and guarded.

If it helps your own testing: the ROCm build needs `RPU_CXX_STANDARD=20` when built against recent ROCm PyTorch wheels (their headers use C++20), and the exact CMake invocations are in the PR description. If there is a specific test or workload you want to see run on this hardware, name it and I will post the output.

PabloCarmona:
Hello again @jeffdaily ! We are not gonna be able to maintain this because we are not working with AMD GPUs right now. But for us is a good addition for the community and for the adoption of the kit for the AMD users. So we are thinking if you can make a fork of this for that specific AMD usage. I don't know what do you think about this. Let us know and we can discuss it further. Thanks again for such great work!

jeffdaily:
We do already have the fork of aihwkit located at https://github.com/AMD-Ecosystem/aihwkit, and the branch this PR is based on https://github.com/AMD-Ecosystem/aihwkit/tree/moat-port.  Though our goal was always to upstream, we do understand that for one reason or another our PRs might not be acceptable by upstream maintainers.  Our fork and this PR branch, if unmerged here, will be long-lived anyway.  Would you be opposed to keeping the PR open so that others might discover it and our fork?  We do have some agent processes that periodically check up on all the projects we've ported, looking for file conflicts or maintainer comments and reviews.  So this PR could be long-lived, as well.  What do you think?

PabloCarmona:
We can add this at our [README](https://github.com/IBM/aihwkit/blob/master/README.md), so the users and community can see that they have a fork available to make the kit work with AMD GPUs natively. This approach sounds good to you?

jeffdaily:
Works for me for your README to point to our fork + branch.  Do you plan to keep this PR open then or no?  The benefit of keeping it open is the provenance and findability if someone comes looking for AMD support, plus if there are file conflicts as your project evolves they're easy to discover.  But I understand if you want your project tidy by closing the PR, and I'll figure out a way to adapt my processes.

PabloCarmona:
We closed this one @jeffdaily and put as you can see a section on the [README](https://github.com/IBM/aihwkit/blob/master/README.md#running-on-amd-gpus-rocmhip), there the people can look also at this PR even if it closed. Thanks for the effort you put to have our toolkit in your AMD ecosystem!

Issue #766, opened 2026-03-26
CUDA test failures on Ampere+ GPUs due to cuDNN TF32 Tensor Core precision


Zhaoxian-Wu:
## Description

Multiple CUDA tests fail on GPUs with TF32 Tensor Core support (Ampere and newer architectures, sm >= 80). The root cause is a precision mismatch between the **digital reference path** (which uses cuDNN with TF32 enabled by default) and the **analog simulation path** (which uses CUBLAS `sgemm` in full FP32 precision).

Since PyTorch 1.7, `torch.backends.cudnn.allow_tf32` defaults to `True`. cuDNN operations on Ampere+ GPUs may internally use TF32 Tensor Cores (19-bit: 8-bit exponent + 10-bit mantissa) instead of full FP32 (32-bit: 8-bit exponent + 23-bit mantissa) for matrix multiplications. This introduces up to ~1e-3 absolute error per GEMM operation. Meanwhile, the RPU C++ backend calls CUBLAS `sgemm` directly, which respects `torch.backends.cuda.matmul.allow_tf32` (default `False`) and thus runs in full FP32. The tests compare these two paths using tight tolerances (`decimal=4` or `decimal=6`), causing failures.

### Affected tests

- `tests/test_layers_convolution.py` - Conv3d CUDA tests (`test_torch_original_layer`)
- `tests/test_layers_rnn.py` - VanillaRNN, GRU, LSTM CUDA tests (`test_layer_training`, `test_bidir_layer_training`)
- `tests/test_layers_rnn.py` - LSTMCombinedWeight CUDA tests (`test_layer_training`)

### Affected hardware

The behavior depends on both GPU architecture and cuDNN's internal algorithm selection heuristic:

| Architecture | Example GPUs | Conv3d (small) | RNN/LSTM/GRU |
|---|---|---|---|
| Volta / Turing (sm_70/75) | V100, T4, RTX 2080 | pass | pass |
| Ampere (sm_80/86) | A100, RTX 3090 | untested | likely fail |
| Hopper (sm_90) | H100 | **pass** | **fail** |
| Blackwell (sm_100/120) | B200, RTX PRO 6000 | **fail** | **fail** |

Note: Conv3d passes on H100 because cuDNN decides the equivalent GEMM (M=3, K=128, N=630) is too small for Tensor Cores on Hopper, but Blackwell has a lower activation threshold for Tensor Cores. This makes the issue architecture-dependent and not simply a function of TF32 hardware support.

[...]

PR #764, opened 2026-03-24
[feature] Non-zero gamma support in ChoppedTransferCompound

Zhaoxian-Wu:
# Non-zero gamma support, `scale_fast_lr`, and documentation

## Overview

This PR makes two improvements to `ChoppedTransferRPUDevice` / `ChoppedTransferCompound`:

1. **Non-zero `gamma` support** — `ChoppedTransferCompound` can now be used as a residual-learning device where the fast array A contributes directly to the effective weight, with correct chopper de-correlation applied during weight reduction.
2. **`scale_fast_lr` parameter** — a new parameter, analogous to the existing `scale_transfer_lr`, that controls whether the fast-device LR tracks the current optimizer LR.
3. **Document update** —  A new paragraph to discuss the benefit of `gamma` is attached to the documents. To provide sufficient context, this PR also expands the algorithm documentation to cover TTv1 through TTv4.

---

## 1. Non-zero `gamma` support in `ChoppedTransferRPUDevice`

Previously, `checkSupported()` enforced `fullyHidden()`, which hard-blocked any configuration where the fast array A contributes to the visible weight (`gamma != 0`). This restriction is lifted, and correct behaviour is implemented via a `reduceToWeights` override (CPU + CUDA).

**Background:** A is updated with per-element chopper sign flips and is therefore stored in "chopped" form: `A_stored[i,j] ≈ c_d[i]·c_x[j]·A_true[i,j]`. The base-class weight-reduction GEMV computes `W = gamma·A_stored + C`, which is incorrect when `gamma != 0` because the chopper factors are not cancelled. The new override applies a correction after the GEMV:

```
W[i,j] += gamma * (c_d[i]*c_x[j] - 1) * A_stored[i,j]
```

The `- 1` term accounts for the fact that the base GEMV already contributed `gamma * A_stored`; the correction adds only the remaining `gamma * (c_d[i]*c_x[j] - 1) * A_stored` to reach the correct `gamma * (c_d[i]*c_x[j]) * A_stored`. On CUDA this is implemented as the new `kernelApplyChopperCorrectionToWeights` kernel; on CPU it is a simple loop. Both paths are no-ops when `gamma == 0` (the default).

[...]

PR #771, opened 2026-06-08
fix(tests): dynamic tolerance for cuDNN TF32 precision on Ampere+ GPUs

Zhaoxian-Wu:
## Summary

This PR addresses issue #766, the CUDA test failures on Ampere+ GPUs caused by the precision mismatch between cuDNN's default TF32 Tensor Core path and the RPU backend's FP32 CUBLAS path.

Instead of globally disabling TF32, this follows the hardware-adaptive direction discussed in the issue:
- add cached CUDA probes for Conv3d and RNN numerical divergence
- derive `assert_array_almost_equal` decimal tolerances from the measured divergence
- keep CPU tests at the existing strict `decimal=6`
- relax CUDA tolerances only when the current GPU/cuDNN behavior requires it

## Testing

The affected CUDA tests now pass on my setup, including the previously failing RNN/LSTM and Conv3d cases on TF32-capable GPUs.

---
Hi @PabloCarmona, this PR helps me pass the full test suite locally. If the tests on the Git CLI pass as well, I will update the code in my other PRs accordingly. Thanks for your effort in maintaining the library.

### DIFF
--- a/tests/helpers/testcases.py
+++ b/tests/helpers/testcases.py
@@ -6,7 +6,9 @@
 
 """TestCases for aihwkit tests."""
 
+import math
 import os
+from functools import lru_cache
 from typing import Type
 from unittest import SkipTest, TestCase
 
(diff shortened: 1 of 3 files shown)


PabloCarmona:
Thanks @Zhaoxian-Wu for your work! Alongside this one, look at the others ones and sync them with master, so the actions triggers again and should be working now since I added the fix to some of the lint errors were arising in the past.

Also pass in the linting tool to address this:
```bash
tests/helpers/testcases.py:56:53: E261 at least two spaces before inline comment
tests/helpers/testcases.py:58:55: E261 at least two spaces before inline comment
```

Zhaoxian-Wu:
Got it. I forgot to test the pycodestyle. This commit should be okay.  Could you please trigger the test again to see whether it's correct? @PabloCarmona

PabloCarmona (approved):
Seems ok for me, please update rest of PRs with the tests accordantly and we can start to finally merge them. I supposed this doesn't need to be merged once the other PRs has its test cases updated.

Zhaoxian-Wu:
Sure, I've updated

PabloCarmona:
Hello @Zhaoxian-Wu, do we need to merge this finally? Or we wait until you made the update in the correspondant PR? I already merge some of them that have tests attached to it, but the https://github.com/IBM/aihwkit/pull/764 don't have any.

Thanks for your incredible work!

PR #785, opened 2026-08-06
fix(optim): support like factories for placeholder contexts

Zhaoxian-Wu:
## Why

`AnalogContext` exposes placeholder data by default so public tensor operations cannot accidentally read meaningless backing weights. Tensor `*_like` factories only need metadata such as shape, dtype, and device, and should be valid in this mode. Before this change, they were rejected as value-reading operations.

This is related to the placeholder-data feature introduced by [IBM/aihwkit#765](https://github.com/IBM/aihwkit/pull/765), and completes its compatibility with standard PyTorch `*_like` factories.

## Failure case

```python
model = AnalogLinear(1, 1, bias=False, rpu_config=SingleRPUConfig(
    device=ConstantStepDevice()
)).to("cuda")
torch.zeros_like(ctx)
```

Before this fix, the final line raised:

```text
RuntimeError: AnalogContext data is in placeholder mode, so operation 'zeros_like' cannot read weight values.
```

This prevented metadata-only tensor construction during CUDA model setup and other generic PyTorch code paths.

## What changed

- Allow `empty_like`, `full_like`, `ones_like`, `rand_like`, `randint_like`, `randn_like`, and `zeros_like` in placeholder mode.
- Keep the placeholder backing storage unread while allowing PyTorch to allocate the result from metadata.
- Add CPU coverage for all seven public `*_like` factories and a CUDA regression test for `zeros_like`.

### DIFF
--- a/src/aihwkit/optim/context.py
+++ b/src/aihwkit/optim/context.py
@@ -21,6 +21,7 @@ from aihwkit.optim.weight_view import (
     raise_if_readonly_write_target,
     PlaceholderDataView,
     _PLACEHOLDER_METADATA_FUNCTIONS,
+    _PLACEHOLDER_LIKE_FACTORY_FUNCTIONS,
     _raise_placeholder_read_error,
 )
 from aihwkit.simulator.parameters.enums import AnalogContextDataViewMode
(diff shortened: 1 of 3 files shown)


Zhaoxian-Wu:
Hi @PabloCarmona,
All the test passed in my local environment. However, the test failed intermittently ([test-and-lint / Test 3.12](https://github.com/IBM/aihwkit/actions/runs/31124566023/job/92692423354?pr=785)) because two random noisy outputs happened to quantize to the same value. Could you please review this stochastic test case?

Possible solution: [tests/test_simulator_tiles.py:686](https://github.com/IBM/aihwkit/blob/836cdd8f2be306d225625ae129ff7ff8c7dc904b/tests/test_simulator_tiles.py#L686) compares only one randomly selected noisy-output pair, so it can falsely fail when both samples quantize to the same value.

PabloCarmona:
Thanks @Zhaoxian-Wu ! I will take a look at the PR and what are you talking about the tests for the random noisy value, we are aware of it, we will take a look again.

Thanks again!

PabloCarmona:
@Zhaoxian-Wu could you sync up this branch with the master branch? We are planning to launch a new release very soon. Thanks again!

Zhaoxian-Wu:
Hi @PabloCarmona. Sure thing. I've pull the branch and it's undergoing test. Hopefully it would work this time.

PabloCarmona (approved):
LGTM!

PR #786, opened 2026-08-21
Fix double bracket indexing bug on Hermes noise model

atvasilopoulos:
## Description

Double brackets were used to index a torch tensor in the hermes noise model, leading to a UserWarning when using it. Fixed this by changing it to proper indexing.

## Details

The following warning was raised when using the hermes noise model (`model.program_analog_weights()`). 

```
UserWarning: Using a non-tuple sequence for multidimensional indexing is deprecated and will be changed in pytorch 2.9; use x[tuple(seq)] instead of x[seq]. In pytorch 2.9 this will be interpreted as tensor index, x[torch.tensor(seq)], which will result either in an error or a different result (Triggered internally at /__w/pytorch/pytorch/torch/csrc/autograd/python_variable_indexing.cpp:349.)
  g_relative[[g_relative >= 0.0945]],
```

Found 5 instances where this indexing pattern was used in the class. Single bracket solves the issue.


PabloCarmona:
Hey @atvasilopoulos ! Could you add the authoring sign in to your commits so the DCO check passes? Thanks!

PabloCarmona (approved):
LGTM!

PR #791, opened 2026-09-11
fix(uv.lock): solve vulnerability dep review for accelerate

PabloCarmona:
## Related issues

https://github.com/IBM/aihwkit/security/dependabot/58

## Description

Solve vulnerability found for accelerate dependency.

## Details

- Sync [uv.lock](https://github.com/IBM/aihwkit/blob/master/uv.lock) file.


PabloCarmona:
Closes https://github.com/IBM/aihwkit/issues/790

